PR #5, opened 2019-06-26
tools: add script to display MCU firmware footprint

Yanok35:
This display graphbar based on available Flash & RAM in the MCU
and after checking GNU size outputs of a freshly built firmware.

Signed-off-by: Yannick Gicquel <ygicquel@gmail.com>

### DIFF
--- a/boards/cogip2019-cortex/Makefile.include
+++ b/boards/cogip2019-cortex/Makefile.include
@@ -2,6 +2,12 @@
 export CPU = stm32f4
 export CPU_MODEL = stm32f446re
 
+# compute FLASHSIZE and RAMSIZE
+include $(RIOTCPU)/stm32_common/stm32_mem_lengths.mk
+KB := 1024
+LEN := $(shell echo $(RAM_LEN) | sed 's/K//')
+RAMSIZE := $(shell echo $$(( $(LEN) * $(KB) )) )
+
 export DEBUG_ADAPTER ?= stlink
 export STLINK_VERSION ?= 2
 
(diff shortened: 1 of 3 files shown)


boards/cogip2019-cortex/Makefile.include
```diff
@@ -1,6 +1,8 @@
 # define the cpu used by the cogip2018-cortex board
 export CPU = stm32f4
 export CPU_MODEL = stm32f446re
+export CPU_FLASH_SIZE = $(shell expr 512 \* 1024)
```
gdoffe: This should be setup into Makefile.include in mcu-firmware base folder.
Some variables already exist from cpu/stm32_common/stm32_mem_lengths.mk and example of use can be found into cpu/stm32_common/Makefile.include.

- https://github.com/RIOT-OS/RIOT/blob/master/cpu/stm32_common/stm32_mem_lengths.mk
- https://github.com/RIOT-OS/RIOT/blob/a01e70a4ab0050331b36ad59946637c62c48ed3a/cpu/stm32_common/Makefile.include#L22

It will make code generic and will lighten platform Makefile

platforms/cogip2019-cortex/Makefile
```diff
@@ -41,3 +41,10 @@ CFLAGS += -DTHREAD_STACKSIZE_MAIN=4096
 DEVELHELP=1
 
 include ../../Makefile.include
+
```
gdoffe: This should be done into Makefile.include in mcu-firmware base folder.

boards/cogip2019-cortex/Makefile.include
Yanok35: Was not aware, i'll take a look . thx

PR #6, opened 2019-10-03
Makefiles refactoring

gdoffe:
Concerning makefiles, the platform management was a mess with duplicate entries and a bad common stuff management.

This pull request intends to correct that by introducing the MCUFIRMWARE_PLATFORM_BASE variable that contains the name of the common base platform code to use.

For instance, MCUFIRMWARE_PLATFORM_BASE set to cortex-common will include platforms/common/cortex-common/Makefile.include to the platform Makefile.

Moreover, the 'all' target is now available to build all existing platforms.


Makefile
```diff
@@ -1,4 +1,10 @@
-.PHONY: all doc doc-man doc-latex docclean
+.PHONY: all allpf doc doc-man doc-latex docclean pfs
```
Yanok35: allpf does not exist
target "doc-latex" is named "latex" below
pfs is a variable, i think there is no need to declare it as a phony target. maybe you were looking to $(pfs) expression evaluation at this point?

Makefile
```diff
@@ -1,4 +1,10 @@
...
+
+# List all platforms
+pfs := $(foreach dir,$(wildcard platforms/*/Makefile),$(subst Makefile, , $(dir)))
+
+all: pfs
```
Yanok35: all: $(pfs) ?

Yanok35:
In 
0f5fbbf, "platforms/cogip-2019-cortex-common/include/platform-common.h" moved to "platforms/common/cortex-common/include/platform-common.h".
New file path contains three times "common" identifier... <joke> seriously ? come on :) </joke>

A "platform" is already defining a "common base" for derivatives projects, isn't it ?
What do you think of a location like "platforms/include/" for headers, and "platforms/" to put sources files for example ?

Anyway. the PR does not build in Travis for some reasons.

Makefile
gdoffe: Yes i'm not even sure that this dependency is needed as existing platforms are listed into the variables...

gdoffe:
> In
> [0f5fbbf](https://github.com/cogip/mcu-firmware/commit/0f5fbbfc80c906da2d8fa3559b091aca003933bc), "platforms/cogip-2019-cortex-common/include/platform-common.h" moved to "platforms/common/cortex-common/include/platform-common.h".
> New file path contains three times "common" identifier... < joke > seriously ? come on :) < /joke >

Indeed there is a common that should not be there at least: "platforms/common/cortex-**common**/include/platform-common.h"
However platform-common.h cannot be renamed as we cannot have 2 header files with the same name (platform.h).

> A "platform" is already defining a "common base" for derivatives projects, isn't it ?
> What do you think of a location like "platforms/include/" for headers, and "platforms/" to put sources files for example ?
No a platform is defining the final project. It is containing the main Makefile. It is the start point.
However some properties/functions can be shared between platforms and thus Makefiles too.
So I created a common directory in the RIOT-OS way of life you love so much. :)
The purpose of this is to keep an up to date common "platform" for testing and easy start of a new platform.
For instance creating multiple test platform on a common base according to the subject we could work on.

> Anyway. the PR does not build in Travis for some reasons.
Yes due to external package and to the fact that the .travis.yml is making RIOT pointing to master branch instead to a tag...

Yanok35:
> > In
> > [0f5fbbf](https://github.com/cogip/mcu-firmware/commit/0f5fbbfc80c906da2d8fa3559b091aca003933bc), "platforms/cogip-2019-cortex-common/include/platform-common.h" moved to "platforms/common/cortex-common/include/platform-common.h".
> > New file path contains three times "common" identifier... < joke > seriously ? come on :) < /joke >
> 
> Indeed there is a common that should not be there at least: "platforms/common/cortex-**common**/include/platform-common.h"
> However platform-common.h cannot be renamed as we cannot have 2 header files with the same name (platform.h).
Ok, let's move from "cortex-common" to "cortex".
> 
> > A "platform" is already defining a "common base" for derivatives projects, isn't it ?
> > What do you think of a location like "platforms/include/" for headers, and "platforms/" to put sources files for example ?
> 
> No a platform is defining the final project. It is containing the main Makefile. It is the start point.
> However some properties/functions can be shared between platforms and thus Makefiles too.
> So I created a common directory in the RIOT-OS way of life you love so much. :)
> The purpose of this is to keep an up to date common "platform" for testing and easy start of a new platform.
> For instance creating multiple test platform on a common base according to the subject we could work on.
I am not convinced on the source tree yet, but this may change later :) let's move forward and see how it does...

As Makefile design is changing upstream, i think 
> 
> > Anyway. the PR does not build in Travis for some reasons.
> 
> Yes due to external package and to the fact that the .travis.yml is making RIOT pointing to master branch instead to a tag...

quick fix in the PR#7. Please rebase when merged, this will fix CI build

Makefile
Yanok35: Agree. Looking at the macro expansion, this target will generate multiple sub-make instances. Agree, no need to add target dependencies, this can be handled in sub-make instances.

Makefile
```diff
@@ -1,14 +1,28 @@
-.PHONY: all doc doc-man doc-latex docclean
+.PHONY: all doc docman doclatex docclean help
 
-doc:
+# List all platforms
+pfs := $(foreach dir,$(wildcard platforms/*/Makefile),$(subst Makefile, , $(dir)))
+
+all: 		## Build all platforms
```
Kheprys: Comment should mention it's the default target it would appear on make help command.

Makefile
```diff
@@ -1,14 +1,28 @@
...
+	$(foreach pf, $(pfs), make -j$$(nproc) -C $(pf);)
```
Yanok35: Other documentation related target below uses "$(MAKE)" variable instead of direct call to make.

gdoffe: Ah yes good point !

Makefile
```diff
@@ -1,5 +1,17 @@
 .PHONY: all doc doc-man doc-latex docclean
```
Yanok35: PHONY variable need alignment with new targets introduce by this commit.
Please check https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

Whereas this link highlight the aim of phony rule, it also explain a potential issue while looping over directory to call recursive make as done below.
I checked quickly, but this might apply to this commit as well : if I understand correctly, the issue could be a sub-make failing while encountering a build error, but treated as a successful build by the parent make.

gdoffe: Good catch @Yanok35.
I followed the advices from the link you provided by declaring the sub-directories as .PHONY targets.

As several targets could depend on $(pfs), I used a variable to pass the "sub make" target:
```
distclean: PF_TARGET = distclean
distclean: $(pfs)

$(pfs):
	$(MAKE) -j$$(nproc) -C $@ $(PF_TARGET)
```
https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html

It seems functionnal but what do you think about this ?

PR #7, opened 2019-10-06
Ygl riot stable version

Yanok35:


controllers/include/ctrl.h
```diff
@@ -182,7 +182,7 @@ void ctrl_set_pose_intermediate(ctrl_t *ctrl, uint8_t intermediate);
  *
  * @return                      normal position when 0, intermediate otherwise
  */
-int ctrl_get_pose_intermediate(ctrl_t *ctrl);
+uint8_t ctrl_is_pose_intermediate(ctrl_t *ctrl);
```
gdoffe: Can you please update the doxygen accordingly.
I suggest
```
/**
 * @brief Check if the pose order is an intermediate position
 *
 * @param[in] ctrl              Controller object
 *
 * @return                      non-zero on intermediate position, 0 otherwise
 */
```

Yanok35: done. i just pushed updated branch. thanks

PR #10, opened 2019-11-14
Ygl pid monitoring updates

Yanok35:


gdoffe:
Hi Yannick,

Thanks for all that work.

However it is a bit hard to review due to the great number of commits and to a lack of git messages.

Can you please group and start with the commits related to the controller, followed by platforms commit and simulation at the end ? it should just be a rebase with no conflicts.

Moreover there is a lot of commits without body message:
5485877 (HEAD -> ygl-pid-monitoring-updates, origin/ygl-pid-monitoring-updates) readme: add note on how to launch pid monitoring tool
264232b simu: pid: use getopt to select uart or native parser
66cc56d simu: pid: add few messages
a47ec38 simu: pid: embed main UI in QMainWindow derived class
2eb60a4 simu: pid: update position
a964b6b simu: pid: add speed pid graphs
477320a simu: pid: update Parser and Qt threads to use RobotTelemetry
74b42c3 ctrl: add a cycle counter incremented on each control loop turn
06252fd readme: add note about virtualenv usage
cbc5d23 misc: add python virtualenv directory in gitignore
619b3c7 controllers: store speed_order instead of pointer in instance

Understanding what the commit does is not obvious when not in context. Describing the context and giving few details about what the commit does would help.

Thanks,
Gilles

Yanok35:
Thanks for reviewing. New commit messages freshly pushed.
Not sure about your concern on controller related commits, please tell me if still unclear.

Yanok35:
Well, I think I do not understand the way Github display commits... The commit graph is not identical between my local repository and Github.
Here is start of this PR, seems Github jeopardized them at display ?

$ git log --pretty=oneline --abbrev-commit
(...)
* 57d5dc2 simulation: add requirements.txt
* c9f665e controllers: add a cycle counter incremented on each control loop turn
* beb3159 controllers: store speed_order instead of pointer in instance
* 4a2c878 (origin/master, origin/HEAD) quadpid: remove obsolete pointer check

In Github commit 57d5dc2 & c9f665e are swapped ?! How is it for you ?

gdoffe:
Hi Yannick,

Looks better now, thanks.

I will review tomorrow.

23b85aa (HEAD -> ygl-pid-monitoring-updates, origin/ygl-pid-monitoring-updates) readme: add note on how to launch pid monitoring tool
6fa9a50 simu: pid: use getopt to select uart or native parser
b622d3f simu: pid: add doc attribute to Parser object
997e9b2 simu: pid: add a toolbar to GraphWindow
18a9e40 simu: pid: add GraphWindow object
1fef044 simu: pid: propagate robot pose in MapView
c84ad39 simu: pid: add MapView object
5b8c800 quadpid: calib: change getchar expected character
db149cc simu: pid: add speed pid graphs
03ecbb6 simu: pid: update Parser and Qt to use RobotTelemetry
3d5ef23 simu: pid: add RobotTelemetry object
b347e4d simu: freecad: update parser indexes
71ebf4b all: move control loop measurement point
15ad536 simu: pid: add CurveOverTime object
95000e1 simu: pid: remove global array
09664d7 simu: pid: bump script to python3
cdd16b6 simu: pid: add modeline
47da426 readme: add note about virtualenv usage
de32f53 misc: add python virtualenv directory in gitignore
8eb46fb platform: display calibration info message when built-in
57d5dc2 simulation: add requirements.txt
c9f665e controllers: add a cycle counter incremented on each control loop turn
beb3159 controllers: store speed_order instead of pointer in instance

Yanok35:
Hi, please note I have seen a bug on tooltip display I just fixed and force-pushed few minutes ago. Before diving in all the commits, please fetch the latest updates. I won't touch the branch anymore, unless someone see something to change. Thanks !

controllers/quadpid/quadpid.c
```diff
@@ -142,26 +148,13 @@ int ctrl_quadpid_running_speed(ctrl_t* ctrl, polar_t* command)
...
-                pose_current->x,
-                pose_current->y,
-                pose_current->O);
-
-    DEBUG("@robot@,%u,@speed_current@,%.4f,%.4f\n",
-                ROBOT_ID,
-                speed_current->distance,
-                speed_current->angle);
-
-
```
gdoffe: I totally disagree with this move outside of the controller for several reasons:

1) If we want to debug a controller, it is to the controller to provide all the logs required to its debug. I should not need to set ENABLE_DEBUG in platforms and in odometry for this.
2) if I change odometry for IMU for example, I will have to check that they expose logs too to ensure that external tools like simulation are still working.
3) In addition to the several ENABLE_DEBUG I need to activate, log function search or development will be harder if splitted into several files.

What could be done is to add filter keywords with the displayed string:
Instead of "@robot@" we could have "@robot,ctrl@", "robot,ctrl,quadpid@", "@robot,odometry@"

The scope would be clear and if for some reasons logs could be duplicated because of using the same information they could be filtered easily by external tools.
If ENABLE_DEBUG for odometry is not set thus it will not display duplicate information if not needed.

So please lets keep controller stuff into controller scope.

platforms/common/cortex/platform-common.c
```diff
@@ -119,6 +142,19 @@ void motor_drive(polar_t *command)
     int16_t right_command = (int16_t) (command->distance + command->angle);
     int16_t left_command = (int16_t) (command->distance - command->angle);
 
+    ctrl_t *ctrl = pf_get_ctrl();
+
+    if (ctrl_get_mode(ctrl) != CTRL_MODE_STOP) {
```
gdoffe: and different of  CTRL_MODE_IDLE ?
and maybe CTRL_MODE_BLOCKED

simulation/robot_calibration.py
```diff
@@ -30,6 +30,12 @@
 win.setWindowTitle("PID loop monitor")
 win.resize(1080, 720)
 
+def deg2rad(deg):
+    return deg * 2. * math.pi / 360.
+
+def rad2deg(rad):
+    return rad * 360. / (2 * math.pi)
+
```
gdoffe: As you import math with not using builtin functions ?
https://docs.python.org/2/library/math.html#angular-conversion

simulation/robot_calibration.py
```diff
@@ -95,6 +101,71 @@ def clear(self):
         self._addDataItems()
 
 
+class MapView(pg.PlotItem):
+    """
+    This object allow a representation of the robot map.
+
+    It draw a fixed rectangle for the game area, according to firmware's frame
```
gdoffe: draws

simulation/robot_calibration.py
```diff
@@ -95,6 +101,71 @@ def clear(self):
         self._addDataItems()
 
 
+class MapView(pg.PlotItem):
+    """
+    This object allow a representation of the robot map.
```
gdoffe: allows

README.md
```diff
@@ -48,6 +48,16 @@ Edit ~/.bashrc file and add $HOME/toolchain/gcc-arm-none-eabi-8-2018-q4-major/bi
...
+# Prepare python environment
+
+Some tool rely on python to run. A possible preparation is proposed below:
+
+```bash
+$ virtualenv simulation/venv
+$ source simulation/venv/bin/activate
+$ pip3 install -r simulation/requirements.txt
+```
+
```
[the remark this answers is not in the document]
ecourtois: Please mention that Python>=3.6 is required to run the simulation script.
The right way to create a virtualenv since Python 3.4 is the following:
```suggestion
$ python3 -m venv simulation/venv
$ source simulation/venv/bin/activate
$ pip install -r simulation/requirements.txt
```
This way, only python3 and python3-venv package are required.
After venv activation, python and pip version defaults to the version used to create the venv, so you can juste run 'python' and 'pip' instead of 'python3' and 'pip3'.

README.md
```diff
@@ -62,6 +72,23 @@ $ make -j$(nproc)
...
 ```
 
+## Build and observe control loop
+
+A python script is available to observe the data of the control loop.
+Script can be launched using following:
+
+```bash
+$ source simulation/venv/bin/activate
+$ python3 simulation/robot_calibration.py
```
ecourtois:
```suggestion
$ python simulation/robot_calibration.py
```

README.md
```diff
@@ -48,6 +48,16 @@ Edit ~/.bashrc file and add $HOME/toolchain/gcc-arm-none-eabi-8-2018-q4-major/bi
...
 ```
 
+# Prepare python environment
+
+Some tool rely on python to run. A possible preparation is proposed below:
+
+```bash
+$ virtualenv simulation/venv
+$ source simulation/venv/bin/activate
+$ pip3 install -r simulation/requirements.txt
```
ecourtois: Please mention that Python>=3.6 is required to run the simulation script.
The right way to create a virtualenv since Python 3.4 is the following:
```
$ python3 -m venv simulation/venv
$ source simulation/venv/bin/activate
$ pip install -r simulation/requirements.txt
```
This way, only python3 and python3-venv package are required.
After venv activation, python and pip version defaults to the version used to create the venv, so you can juste run 'python' and 'pip' instead of 'python3' and 'pip3'.

simulation/robot_calibration.py
```diff
@@ -141,101 +416,157 @@ def _decode(self, line):
...
+    uart_dev = None
+    for o, a in opts:
+        if o in ("-h", "--help"):
+            usage(pname = sys.argv[0])
+            sys.exit()
+        elif o in ("-D", "--device"):
+            uart_dev = a
+        else:
+            assert False, "unhandled option"
+
```
ecourtois: getopt module is C-developers friendly but almost deprecated. Please consider using [argparse](https://docs.python.org/3.7/library/argparse.html#module-argparse) module instead.

README.md
Yanok35: Ok, good to know. This will be updated accordingly.

controllers/ctrl.c
```diff
@@ -117,6 +117,9 @@ void ctrl_set_mode(ctrl_t* ctrl, ctrl_mode_t new_mode)
     if (new_mode < CTRL_MODE_NUMOF) {
         ctrl->control.current_mode = new_mode;
         DEBUG("ctrl: New mode: %d\n", ctrl->control.current_mode);
+
+        if (ctrl->control.current_mode == CTRL_MODE_STOP)
```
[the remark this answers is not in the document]
Yanok35: Maybe this would make sense to always set current_cycle to 0.
Anyway this feature is used for 1) multi-line log aggregation, 2) controller speed order pattern generation which are function of times (not part of this patchset)... need to give a try to check if no regression without this test.

controllers/quadpid/quadpid.c
Yanok35: Except for displaying numerical values, "pose_current" is not even used in this function which regulates speed only. Code to display this numerical information was factorized.

Anyway, your comments seems related to the way logs are enabled per file in the overall project and the way the controller periodic loop was spread in multiple files, which is out of scope of the proposed work.

>What could be done is to add filter keywords with the displayed string:
>Instead of "@Robot@" we could have "@Robot,ctrl@", "robot,ctrl,quadpid@", "@Robot,odometry@"

This makes me remind a critical point 'totally' dropped during previous refactoring activities : the current implementation is not considering the time overhead introduced by the UART, which is an issue.

Initial implementation was using a RAM based buffer (log_vect) service to first catch numerical value of a cycle between one measurement and one command, and secondly to print those values _after_ command was set, ensuring a minimum time overhead of the numerical value extraction on the controller performance.

https://github.com/cogip/avr-firmware/blob/3e9ed4ebe72fe4e7ad636e01b70284d6e1adf7ec/drivers/system/log.h#L44

In the current design, the overall probes points are adding extra delay which are nearly consuming the 20ms timeframe (around 230 chars printed by uart * 86.4us ~= 19.8 ms), thus adding an extra delay of almost one controller cycle... which is nearly breaking the real-time constraints.

platforms/common/cortex/platform-common.c
Yanok35: This test is to avoid log flood when mobile robot feedback loop is 'not-running'.
Goal is to keep users interactive session through UART usable feature.
CTRL_MODE_IDLE is currently dead code, but agree this should be added in case it appears in the object initialization implementation.
CTRL_MODE_BLOCKED, even if it's a transient mode, could produce data on wheels and robot movement which could be interesting, good to keep if can be probed.

platforms/common/cortex/platform-common.c
```diff
@@ -364,19 +364,21 @@ void pf_init_tasks(void)
...
+
     /* Create thread that up a flag on key pressed to start a shell instead of
        planner below */
     kernel_pid_t start_shell_pid = thread_create(start_shell_thread_stack,
                   sizeof(start_shell_thread_stack),
                   THREAD_PRIORITY_MAIN + 1, 0,
                   pf_task_start_shell, &start_shell, "shell");
 
-    DEBUG("Press Enter to enter calibration mode...\n");
+    printf("Press Enter to enter calibration mode...\n");
```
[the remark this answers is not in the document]
Yanok35: Ok, will be done.

platforms/common/cortex/platform-common.c
```diff
@@ -364,19 +364,21 @@ void pf_init_tasks(void)
...
 
     /* Wait for Enter key pressed or countdown */
     while ((!start_shell) && (countdown > 0)) {
         xtimer_ticks32_t loop_start_time = xtimer_now();
-        DEBUG("%d left...\n", countdown);
+        printf("%d sec. left...\n", countdown);
```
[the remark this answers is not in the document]
Yanok35: Ok.

controllers/ctrl.c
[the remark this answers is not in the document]
gdoffe: I agree

simulation/robot_calibration.py
```diff
@@ -239,3 +239,5 @@ def send(self, p):
 
     if sys.flags.interactive != 1 or not hasattr(pg.QtCore, 'PYQT_VERSION'):
         pg.QtGui.QApplication.exec_()
+
+# vim: tabstop=4 expandtab shiftwidth=4 softtabstop=4
```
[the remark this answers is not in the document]
gdoffe: OK as disgussed together, lets keep the vim modelines as the reference for others editors.

platforms/common/cortex/platform-common.c
gdoffe: I agree

controllers/quadpid/quadpid.c
gdoffe: Yes indeed, log buffering is mandatory. Lets merge this as is and we could enable log filtering later, putting first the effort on log buffering.

simulation/robot_calibration.py
Yanok35: Ok, will update thx

PR #11, opened 2019-11-23
Split platform into application + platform

gdoffe:
First try to split platform into application + platform.

platform = hardware external periphereal
application = what is done with the platform

As the sd21 needed to be refactored as it was including app.h, I optimized the driver at the same time to remove duplicated code, speed parameter and use right i2c functions.

This PR also creates the cogip2020-cortex application.


Yanok35:
Hello Gilles,

Nice start, this looks promising,
Commit 4cca926 to 03ee2d7 (included) looks ok to me.

```
| * 4cb6c2f (origin/gdo_platforms) drivers: sd21: fix calibration calls
| * 107d34d applications: add cogip2020-cortex
| * 0f0462d platforms: move init stuff from app to pf
| * 59d93a8 platforms: cortex: move sensors management from app
| * 4c01072 applications: put path in application folder
| * ff1f059 drivers: sd21: remove dependency on application
| * bf85485 applications: update prefix pf_ to app_
| * 03ee2d7 (HEAD -> gdo_platforms) platforms: cortex: remove app.h include
| * e1c4f53 platforms: add pf_init_quadpid_params()
| * c89a2ce platforms: cortex: move camp macros definitions
| * 1d0a71f platforms: cortex: move robot mechanical parameters
| * f880e72 boards: move motors and encoders macros
| * e533c42 applications: cogip2019-cortex: remove cc110x code
| * 58c66d0 boards: cogip2019-cortex: mpu_stack_guard is board related
| * 4f37303 applications: remove cogip2019-cortex-simulation
| * 26d9e67 travis: remove platforms
| * 4cca926 platforms: move platforms to applications
```

The bf85485 is doing a large namespace conversion and I think part of the definitions should stay in platform (it might be beneficial to split this commit).

About sd21 drivers, I propose it is firstly stubbed to enable build in native architecture.
Other enhancement could be done in a second step:  if not another PR at least in some dedicated commits as you started.

About 4c01072, this makes sense to move some part of path.c in app. Current file implements an object definition and services on "path_t" object, but also store the object instance used by app. Thus I think definition of the path should be move outside the current file, but part of functions handling path_t object could be kept as is.

Since this work is reorganizing both source file and their contents, I would propose to integrate till 03ee2d7 with a fix on the native board build (thus stubbing sd21 I guess), ensuring the new files are there and both targets build as before.
Then, in second step dispatch sources file content in platform and app, enhance sd21, add a new app for 2020... if the overall make sense, considering technical details I may have missed.

gdoffe:
> The [bf85485](https://github.com/cogip/mcu-firmware/commit/bf854852a3b890fc217347dc0f9019d8f64f9521) is doing a large namespace conversion and I think part of the definitions should stay in platform (it might be beneficial to split this commit).

Ack, I will split this commit and give explanations about why it should be in app or pf.

> About sd21 drivers, I propose it is firstly stubbed to enable build in native architecture.
> Other enhancement could be done in a second step: if not another PR at least in some dedicated commits as you started.

Ok this commit could also be splitted : refactoring and removing app.h. I will get it out of the PR.
The i2c stub was already done this week end some I will regroup this in some new PRs.

> About [4c01072](https://github.com/cogip/mcu-firmware/commit/4c0107228a7b4da6e4a7cae2a288a0eceffd2104), this makes sense to move some part of path.c in app. Current file implements an object definition and services on "path_t" object, but also store the object instance used by app. Thus I think definition of the path should be move outside the current file, but part of functions handling path_t object could be kept as is.

Yes, this copy/paste was just a quick fix to avoid wrting one path for all apps. Time to start a real path API and indeed store the related objects in the application.

> Since this work is reorganizing both source file and their contents, I would propose to integrate till 03ee2d7 with a fix on the native board build (thus stubbing sd21 I guess), ensuring the new files are there and both targets build as before.

Ok for me. I will do that ASAP.

Kheprys:
The time as come. The merge is finally hapening !

PR #16, opened 2020-08-14
optimizations

gdoffe:
Two enhancements:

* introduce simulation/real configurations
* remove `#ifndef/#define` headers protection macro by `#pragma once`

### DIFF
--- a/applications/cogip2019-cortex/Makefile
+++ b/applications/cogip2019-cortex/Makefile
@@ -19,5 +19,7 @@ USEPKG += vl53l0x-api
 
 # Main thread stack size
 CFLAGS += -DTHREAD_STACKSIZE_MAIN=4096
+# Number of positions available for each servomotor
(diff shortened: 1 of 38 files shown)


applications/cogip2019-cortex/Makefile
```diff
@@ -18,6 +18,6 @@ FEATURES_REQUIRED += periph_i2c
 USEPKG += vl53l0x-api
 
 # Main thread stack size
-CFLAGS += -DTHREAD_STACKSIZE_MAIN=4096
+CFLAGS += -DTHREAD_STACKSIZE_MAIN=4096 -DSD21_SERVO_POS_NUMOF=3
```
Kheprys, line 21: Maybe should be great to add a comment in sd21.h to tell that this NUMOF is in the makefile and not in the code.

gdoffe: done

PR #19, opened 2020-08-20
Add a shell command to list dynamic obstacles.

ecourtois:


robotics/avoidance.c
```diff
@@ -371,3 +371,34 @@ pose_t dijkstra(uint16_t target, uint16_t index)
 dijkstra_error_no_destination:
     return start_position;
 }
+
+int avoidance_print_dyn_obstacles(int argc, char **argv)
```
gdoffe, line 375: I would add this function into obstacle.c as if we develop more avoidance algorithms, we will have to duplicate this code each time.

gdoffe (changes requested):

robotics/avoidance.c
ecourtois: I'd like it too, but the list of polygons is stored in avoidance, and I don't want to refactor this in this PR...

gdoffe: Indeed ! Lets merge it as is and solve this probem in a new PR.

gdoffe (approved):

PR #21, opened 2020-08-23
Add communication between simulator and native firmware

ecourtois:
Use a shared memory array containing sensors values provided by the simulator.
Also add pose order info in output of "_pose" shell command.

### DIFF
--- a/boards/cogip2019-cortex-native/vl53l0x.c
+++ b/boards/cogip2019-cortex-native/vl53l0x.c
@@ -17,10 +17,14 @@
 
 /* Standard include */
 #include <stdio.h>
+#include <sys/shm.h>
 
 /* Project includes */
 #include "vl53l0x.h"
 #include "board.h"
+#include "platform.h"
+
+uint16_t *shm_ptr = NULL;
 
 int vl53l0x_init_dev(vl53l0x_t dev)
 {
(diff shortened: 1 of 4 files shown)


gdoffe (changes requested):
Remove 'Fix' at the end of the commit message

gdoffe (changes requested):
platforms: cortex: add "_set_shm_key" command

boards/cogip2019-cortex-native/vl53l0x.c
```diff
@@ -49,8 +53,20 @@ void vl53l0x_init(void)
...
 uint16_t vl53l0x_continuous_ranging_get_measure(vl53l0x_t dev)
 {
-    (void) dev;
+    /* Try to initialize shared memory if not already done */
+    if(shm_ptr == NULL && pf_shm_key != 0) {
+        int shmid = shmget(pf_shm_key, VL53L0X_NUMOF*sizeof(uint16_t), 0);
+        shm_ptr = (uint16_t*) shmat(shmid,(void*)0,0);
+    }
+
+    /* Return max value is shared memory not initialized */
```
gdoffe, line 62:
```suggestion
    /* Return max value if shared memory is not initialized */
```

gdoffe (changes requested):
Attach to **a** shared memory array

gdoffe (commented):
shell: platform: add pose order to "_pose" command

Add pose order information to the output of "_pose" command.
Used by the simulator to display the current pose order.

gdoffe (changes requested):
**platforms: cortex:** add pose order to "_pose" command

Add pose order information**s** to the output of "_pose" command.
Used by the simulator to display the current pose order.

boards/cogip2019-cortex-native/vl53l0x.c
ecourtois: Attach to **the** shared memory array

gdoffe (approved):

PR #24, opened 2021-01-02
Rename and improve shell command to get robot state

ecourtois:
Rename "_pose" shell command into "_state" as it prints more than positions.
Add cycle and speed info to "_state" command output.
Fix uint32_t printf format specifier.


gdoffe (approved):
LGTM ! Thanks !

PR #25, opened 2021-01-28
Uncrustify update

Kheprys:
Update check-codingrules script to only check if file are compliant with config or not.

It then display diff between uncrustified generated files and original to display what is missing.

Also add a new makefile target 'make check-codingrules' and add this step to github workflow


gdoffe (commented):
tools: update uncrustify coding rules

Update config to stuck C pointer * on type instead of var name.

gdoffe (commented):
tools: update tools/check-codingrules.sh

Only check if source file are compliant and display diff

* add new make command 'make check-codingrules'
* update git workflow to check codingrules

gdoffe (commented):
According to team concertation, we should drop this commit to keep pointer star stucked to the type.

ecourtois (approved):
LGMT

ecourtois (changes requested):
Can you document in README.md how to run this check manually?

Kheprys:
> Can you document in README.md how to run this check manually?

Done !

ecourtois (approved):
LGTM

PR #26, opened 2021-01-29
Fix fetch workflow

Kheprys:
Fix 404 errors from `apt-get install`

Tested 5 times, worked 5 times, seems ok.


gdoffe (commented):
git-workflow: add `apt-get update` to fix 404 errors

Add an `apt-get update` step before installing dependencies to avoid packages to not be
found when issuing `apt-get install` commands.

gdoffe (approved):

PR #28, opened 2021-02-13
Add shell_menu module

ecourtois:
Add a sys directory in mcu-firmware.
Add a first module shell_menu.
This module extends the shell module by providing an API
to declare sub-menus and navigate through them.

Also add an example directory to add example of modules usage.


gdoffe (approved):

Kheprys (approved):
LGTM

PR #32, opened 2021-02-26
Makefiles refactoring

gdoffe:
Big makefiles refactoring to clean and be more consistent on our working way.
This follow RIOT guidelines which gives us more flexibility by turning all software components into modules.
Note: a bug has been spotted in RIOT, I did the fix, thus our own RIOT must be used for now.


gdoffe (commented):
Please say why.

gdoffe (commented):
I love this callback, however say why you add it.
Here you just block automatic path pose increment on pose reached and you start the planner.
You do not initialize it.

Commit title always starts by a verb:
planners: shell: add a menu enter callback

platforms/cortex/platform.c
```diff
@@ -265,18 +266,62 @@ static void *pf_task_countdown(void *arg)
...
+        printf("%d seconds left...\n", countdown);
+        countdown--;
+        xtimer_periodic_wakeup(&loop_start_time, US_PER_SEC);
+    }
+
+    /* Stop task_planner_start_cancel thread if still running */
+    thread_t *planner_start_cancel_thread = (thread_t *)thread_get(planner_start_cancel_pid);
+    if (planner_start_cancel_thread) {
+        sched_set_status(planner_start_cancel_thread, STATUS_STOPPED);
+    }
```
gdoffe, line 324: Just a comment for later but I think this should be placed in a dedicated module.

planners/astar/planner.c
```diff
@@ -78,11 +78,12 @@ static int trajectory_get_route_update(ctrl_t* ctrl, const pose_t* robot_pose,
             /* If the targeted pose is the current path pose to reach, launch
              * the action and update the current path pose to reach to the next
              * one in path, if allowed. */
-            DEBUG("planner: Controller has reach final position.\n");
+            DEBUG("planner: Controller has reached final position.\n");
 
-            /* If it exists launch action associated to the point and increment
-             * the position to reach in the path */
+            /* If not in calibration mode */
```
gdoffe, line 83: shell mode ? (check commit message too)

planners/astar/planner.c
```diff
@@ -93,14 +94,19 @@ static int trajectory_get_route_update(ctrl_t* ctrl, const pose_t* robot_pose,
             }
             /* Update current path targeted position in case it has changed */
             current_path_pos = path_get_current_path_pos(path);
+
             /* As current path pose could have changed, avoidance graph needs
-             * to be recomputed */
+                * to be recomputed */
```
gdoffe: not aligned

planners/astar/planner.c
```diff
@@ -93,14 +94,19 @@ static int trajectory_get_route_update(ctrl_t* ctrl, const pose_t* robot_pose,
...
+            avoidance_update = TRUE;
+        }
+        /* If in calibration mode and new pose is selected */
```
gdoffe, line 102: shell

planners/astar/planner.c
```diff
@@ -93,14 +94,19 @@ static int trajectory_get_route_update(ctrl_t* ctrl, const pose_t* robot_pose,
...
+        else if ((!allow_change_path_pose) && (ctrl_is_pose_intermediate(ctrl))) {
```
gdoffe, line 103: if we have reach an intermediate position it should be `avoidance_index++;`, not `avoidance_update = TRUE;`.

Thats why I deleted it in a previous commit. What do you think ?

gdoffe (commented):
Forgot calibration, it is shell now :P

gdoffe (commented):
Dont you think we should activate uncrustify in the CI now ?
It would avoid this kind of huge commit in the future. ;)

platforms/cortex/platform.c
ecourtois: I agree and I thought about that too, it could be generic, maybe later.

ecourtois:
Yes we should activate uncrustify in the CI.

gdoffe:
> Yes we should activate uncrustify in the CI.

@Kheprys could you do that please ? maybe in a new branch after merge.

planners/astar/planner.c
ecourtois: I've just restored the former code that was working, but I don't clearly understand the planner behavior...

planners/astar/planner.c
ecourtois: Done

planners/astar/planner.c
ecourtois: Done

planners/astar/planner.c
ecourtois: The diff is weird, but the code is aligned...

ecourtois:
Done, restored it from a previous commit!

ecourtois:
`make check-codingrules` is already executed in CI, but result is not check, and it should be now.

gdoffe:
LGTM, just the planner patch I sent you on discord to test. If it works fine for you too. you can merge.

lib/path/include/path.h
```diff
@@ -6,7 +6,7 @@
 typedef struct {
     pose_t pos;
     uint8_t allow_reverse;
-    uint8_t max_speed;
+    double max_speed;
```
gdoffe, line 9: You forgot to say why you did this in the commit message.

PR #33, opened 2021-03-02
Add shmem module and global commands to shell_menu

ecourtois:
sys: shell_menu: Add the possibility to set global commands that will be used in all menus.

sys: shmem: Add a module to help mapping shared memory segments declared by others applications like a simulator.


ecourtois:
Commits have been merged by PR #32

PR #37, opened 2021-04-15
LDS01 DMA Driver: Add native support

ecourtois:
Support native compilation of lds01_dma driver.
Data can be updated using shared memory is configured.

Add the possibility to invert distance and intensity data.

Add configuration of the minimum intensity required to validate a distance.


gdoffe (approved):
LGTM

PR #40, opened 2021-05-02
Add tracefd module

ecourtois:
This module provides an API to write trace files on sdcard.

Since it may be difficult to determine when to close the  files to effectively flush data on disk, and to avoid opening/closing files around each printf, this module starts a thread that periodically closes/opens files to flush data.
This thread can be disabled to allow manual handling of opening/closing files.
Using the lock/unlock API, it is possible to execute consecutive prints in thread safe manner.
The module relies on 'sdcard_spi' driver, so SPI parameters must be provided in the board/platform/application.
Native boards are also supported, files being written in '/tmp'.

This PR also adds support support for remote  flashing and monitoring.


sys/tracefd/tracefd.c
```diff
@@ -0,0 +1,229 @@
...
+/* Declared as `extern` in `RIOT/pkg/fatfs/fatfs_diskio/mtd/mtd_diskio.c` */
+mtd_dev_t *fatfs_mtd_devs[FF_VOLUMES];
+
+/* Define the vfs mount point */
+static vfs_mount_t flash_mount = {
+    .fs = &fatfs_file_system,
+    .mount_point = TRACEFD_ROOT_DIR,
+    .private_data = &fs_desc,
+};
+#endif /* BOARD_NATIVE */
```
gdoffe, line 97: Why is it needed ?

sys/tracefd/tracefd.c
```diff
@@ -0,0 +1,229 @@
...
+tracefd_context_t tracefd_contexts[TRACEFD_NUMOF];
+
+/* Number of initialized tracefd */
+tracefd_t tracefd_initialized = 0;
+
+/* Files flusher thread PID */
+kernel_pid_t files_flusher_pid = -1;
+
+/* Files flusher thread priority */
+#define FILES_FLUSHER_PRIO (THREAD_PRIORITY_MAIN - 1)
```
gdoffe, line 57: Dont you think it should be set to the lowest possible priority ?
https://doc.riot-os.org/thread__config_8h.html#a196e3293c94df476dd1dd0007013ad14
And should it be configurable TRACEFD_NUMOF ?

gdoffe (changes requested):
mostly LGTM, few fixes awaited.

sys/tracefd/tracefd.c
ecourtois: I removed #ifdef and split this into a common file and files specific to the target.

sys/tracefd/tracefd.c
ecourtois: Done

gdoffe (approved):
LGTM ! Nice PR !

PR #42, opened 2021-05-16
Add intensity_threshold command in examples lds01/lds01_dma 

ecourtois:
Add a command to set the intensity threshold.
Under this threshold, distances are considered invalid.


Kheprys (approved):
LGTM

PR #43, opened 2021-05-26
Add trace thread

ecourtois:
* Module sys/tracefd: add stderr and stdout descriptors.
* Module sys/shell_menu: add function to rename a command.
* Use tracefd to print logs and data in JSON format in all modules and applications.
* In applications, add a thread to print robot state automatically on stdout and sdcard, once activated.
* Add a command 'trace_on/trace_off' to activate/deactivate tracing on stdout.
* Makefiles: disable shell prompt and echo to avoid confusing the simulator.

This branch requires the simulator on branch: https://github.com/cogip/simulation/tree/eco-add-replay


sys/tracefd/include/tracefd.h
```diff
@@ -45,12 +45,22 @@ extern "C" {
 
 /**
  * @def TRACEFD_NUMOF
- * @brief Number of trace file descriptors
+ * @brief Number of trace file descriptors specific to the application
  */
 #ifndef TRACEFD_NUMOF
-#  define TRACEFD_NUMOF 1
+#  define TRACEFD_NUMOF 0
```
gdoffe, line 51: why changing to 0 by default ?

sys/tracefd/include/tracefd.h
```diff
@@ -64,15 +74,15 @@ extern "C" {
...
+#  define TRACEFD_FLUSH_INTERVAL 50
 #endif
 
 /**
  * @def TRACEFD_THREAD_PRIORITY
  * @brief Files flusher thread priority
  */
 #ifndef TRACEFD_THREAD_PRIORITY
-#  define TRACEFD_THREAD_PRIORITY THREAD_PRIORITY_MIN
+#  define TRACEFD_THREAD_PRIORITY (THREAD_PRIORITY_MAIN - 1)
```
gdoffe, line 85: why not THREAD_PRIORITY_MIN ?

sys/tracefd/include/tracefd.h
```diff
@@ -122,12 +151,24 @@ void tracefd_unlock(const tracefd_t tracefd);
...
+ *        calling this function.
  * @param[in]   tracefd  trace file descriptor
  * @param[in]   format   format string
  * @param[in]   ...      arguments
  */
 void tracefd_printf(const tracefd_t tracefd, const char *format, ...);
 
+/**
+ * @brief Print formatted string in the trace file in a JSON log message.
+ *        The message must not contain '\n' nor '"'.
```
gdoffe, line 164: Can these characters be escaped ?
Is ' allowed ?
If yes a tracefd_strip_jlog to replace these characters and send the json could be useful ? (in the future maybe)

platforms/cortex/trace_utils.c
```diff
@@ -0,0 +1,62 @@
...
+#include "obstacles.h"
+
+/* Trace filename */
+#define TRACE_FILE "trace.txt"
+
+/* Trace file descriptor on sdcard */
+static tracefd_t tracefd_sdcard;
+
+/* Periodic task */
+#define TASK_PERIOD_MS 100
```
gdoffe, line 20: what is the flush interval then ?

platforms/pf_test/trace_utils.c
```diff
@@ -0,0 +1,62 @@
...
+#include "obstacles.h"
+
+/* Trace filename */
+#define TRACE_FILE "trace.txt"
+
+/* Trace file descriptor on sdcard */
+static tracefd_t tracefd_sdcard;
+
+/* Periodic task */
+#define TASK_PERIOD_MS 60
```
gdoffe, line 20: why 60ms here and 100ms before ?

gdoffe (commented):
LGTM to me, just need some answers

sys/tracefd/include/tracefd.h
ecourtois: Because the module can be used without file, just with stderr and stdout descriptors which are always available.

sys/tracefd/include/tracefd.h
ecourtois: Because otherwise the thread doesn't start...

sys/tracefd/include/tracefd.h
ecourtois: Forbidden characters could probably be escaped, but with a double-escape.
' is allowed.
I'm not in favor of automatically trying to escape unsupported characters because it would change the string length,
be time consuming, and not error prone.
So I would advise to pay attention to the strings given in argument of this function.
In any case, the JSON will be wrong but the program will not crash.

platforms/cortex/trace_utils.c
ecourtois: The state is printed every 100ms while the flush is performed every 1s.
It should be 60ms here, will fix that.

platforms/pf_test/trace_utils.c
ecourtois: Because it allows the simulator to be more precise and it has few impacts on performance.

gdoffe (approved):
LGTM

PR #53, opened 2022-01-22
Fix buffer sizes for Protobuf

ecourtois:
Increase size of shell commands and Profobuf serialization buffer to fix segfault with pf_test and cortex platforms.


gdoffe (approved):
LGTM

PR #54, opened 2022-04-04
Various improvements

ecourtois:
Lot of small fixes/improvements, including:
- reduce size of protobuf messages (float -> integer)
- consider all obstacles in avoidance (not only dynamic obstacles)
- move bounding box handling in obstacle class
- base64-encode protobuf output messages
- app_test: add fixed obstacles on related to game 2022
- pf_test: do not start planner automatically, delegate decision to copilot


platforms/pf_test/include/platform.hpp
```diff
@@ -34,6 +34,7 @@
 #include "odometry.hpp"
```
gdoffe, line 34: Who could use that ?

ecourtois: Not sure which commit you are referring to...

Probably `expose uartpb on public interface` ?
If yes, I dropped that commit because it will concern the next PR about uartpb refactoring.

gdoffe:
LGTM !

gdoffe (approved):

PR #57, opened 2022-04-12
Ensure *.proto files are not always rebuilt

gdoffe:
Create a stamp file when *.protoc files are built to not rebuild
python and C bindings at each build, which could be long.

Also add some CI improvments to handle generated files


ecourtois (approved):
LGTM

PR #58, opened 2022-04-18
Remove unused SPI device

ecourtois:
Remove unused SPI device from the boards.
Make sys/tracefd module work without SDCard.


gdoffe (approved):
LGTM ! Thanks !

PR #59, opened 2022-05-10
Add wizard support at startup

ecourtois:
Wizard is a new library can send questions/messages of different types to Copilot and wait for the answer.
This can be used to build a setup wizard run at startup to configure different parameters of the platform/application.


platforms/cortex/include/platforms/Platform.hpp
```diff
@@ -0,0 +1,74 @@
+// Copyright (C) 2021 COGIP Robotics association
```
gdoffe, line 1: 2022

platforms/cortex/platform.cpp
```diff
@@ -36,9 +36,6 @@
 #define ENABLE_DEBUG        (0)
```
gdoffe, line 36: in commit `sys: shell_menu: factorize handling of Protobuf commands`:
Move code handle => Move code **to** handle

applications/app_test/app_uartpb.cpp
```diff
@@ -0,0 +1,34 @@
+#include "app.hpp"
```
gdoffe, line 1: in commit `platforms: move Protobuf messages handler from platform to application`:
all types of Protobuf messages can be received. => all types of Protobuf messages **that** can be received.

gdoffe (approved):
Few typos.
@ecourtois fix & merge without new review. LGTM  ! good job !

PR #61, opened 2022-05-13
introduce vacuum_pump driver

gdoffe:
In 2022 a vacuum pump driver board has been developped to start/stop
pumps and to check if the vacuum circuit is under pressure.
This driver is using 2 GPIOs:
 - one as output to start/stop the pump
 - one as input to check pressure sensor status


ecourtois (approved):
LGTM!

PR #76, opened 2022-07-29
sys: uartpb: use a unique ID to identify each Protobuf message

ecourtois:
Using a unique ID for each Protobuf message allows libraries/applications defining these messages to register a callback function associated with each ID. This way, uartpb is not dependent of applications anymore, which had to declare a Protobuf message gathering all known Protobuf messages.

This unique ID will be sent first, followed by encoded Protobuf message. The unique id can also be sent alone if the message does not required any content (like a reset or trigger message).


sys/uartpb/UartProtobuf.cpp
```diff
@@ -53,49 +54,83 @@ void UartProtobuf::uart_rx_cb(uint8_t data)
...
-            uart_write(uart_dev_, write_buffer_.get_base64_data(), base64_size+1);
+            base64_size = write_buffer_.base64_encode();
+            if (base64_size == 0) {
+                puts("Failed to base64 encode Protobuf serialized message.");
+                success = false;
+            }
+        }
+    }
+    if (success) {
+        uart_write(uart_dev_, (uint8_t *)&uuid, sizeof(uuid_t));
```
gdoffe, line 117: Juste a question: even if the UUID is not handled by the receiver, you send the message.
It is fine in our case as we need performances, however should it be possible to wait for an ACK to send the message ?

sys/shell_menu/Menu.cpp
```diff
@@ -83,7 +79,7 @@ void Menu::enter(void)
         i++;
     }
 
-    COGIP_DEBUG_COUT("Enter shell menu: " << current_menu->name().c_str());
+    COGIP_DEBUG_COUT("Enter shell menu: " << current_menu->name() << std::endl);
```
gdoffe, line 82: This adds an other new line. is this wanted or an oversight to fix my previous PR error around this ?

sys/uartpb/UartProtobuf.cpp
```diff
@@ -53,49 +54,83 @@ void UartProtobuf::uart_rx_cb(uint8_t data)
...
-        if (res > 0 && message_handler_) {
-            message_handler_(read_buffer_);
+
+        // Read uuid
+        ringbuffer_get(&rx_buf_, (char *)&uuid, sizeof(uuid_t));
+        message_length -= sizeof(uuid_t);
+
+        // Check a handler corresponding to the uuid is registered
+        if (message_handlers_.count(uuid) != 1) {
+            std::cout << "Unknown message uuid: " << uuid << std::endl;
```
gdoffe, line 76: Why not using COGIP_DEBUG_COUT ? (same in all the file)

sys/shell_menu/Menu.cpp
ecourtois: I did not see that COGIP_DEBUG_COUT adds a new line. Actually it should not, I missed that in your previous PR...

sys/uartpb/UartProtobuf.cpp
ecourtois: Because COGIP_DEBUG_COUT is for debug messages and this is an error/log message.

sys/uartpb/UartProtobuf.cpp
ecourtois: It could be be done, but I am not in favor of changing the protocol in this PR.
Moreover, in case of data corruption, waiting for a ACK could lead to a dead-lock.
We could consider this change later if current implementation causes some problems in the future.

sys/shell_menu/Menu.cpp
ecourtois: For example, all functions printing JSON does not work anymore...

sys/uartpb/UartProtobuf.cpp
gdoffe: Use COGIP_DEBUG_CERR :wink:

ecourtois: Nope, still a debug macro...

ecourtois: It was a bad idea to add these macros, we should have just wait to add the log module.

sys/shell_menu/Menu.cpp
ecourtois: I have removed the extra endl here for now, but this will have to be handled properly soon.

gdoffe: > For example, all functions printing JSON does not work anymore...

Can you create an issue about this ?

sys/uartpb/UartProtobuf.cpp
gdoffe: Just create few macros not for debug like COGIP_COUT/COGIP_CERR.
It is completly acceptable to me to have this temporary macros.

sys/shell_menu/Menu.cpp
ecourtois: https://github.com/cogip/mcu-firmware/issues/77

sys/uartpb/UartProtobuf.cpp
gdoffe: Or we can simply assume you let it as is until the log module is ready. What do you think ?

ecourtois: Yes I prefer to wait for the log module.

gdoffe (approved):
LGTM assuming few points on log/error/debug messages !
Good job @ecourtois !

sys/uartpb/UartProtobuf.cpp
ecourtois: https://github.com/cogip/mcu-firmware/issues/78

PR #79, opened 2022-08-01
Add RIOT patching mechanism

gdoffe:
All patches available in patches/ directory will be applied on RIOT sources using quilt.
A patched stamp file is created for each patch applied.
In case of patch deletion, launch a make distclean first.

A first patch to add weak attribute to _init() and _fini() functions in preparation of using libstdc++ is also added.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,12 @@ apps := $(foreach dir,$(wildcard applications/*/Makefile),$(subst Makefile, , $(
 examples := $(foreach dir,$(wildcard examples/*/Makefile),$(subst Makefile, , $(dir)))
 boards := $(foreach dir,$(wildcard boards/*),$(subst boards/, , $(dir)))
 
+# List of patch stamp files
+patchstamps := $(addsuffix ed, $(realpath $(wildcard riot-patches/*.patch)))
+
+# Paths
+riot_path := ../RIOT
+
 .PHONY: all clean distclean doc docman doclatex docclean help $(apps) $(examples)
 
 all: $(apps) $(examples)	## Build all applications and examples (default target)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -12,8 +18,16 @@ clean: $(apps) $(examples)	## Clean build for all applications and examples
 
 distclean: PF_TARGET = distclean
 distclean: $(apps) $(examples)	## Clean build and configuration for all applications
+	@cd $(riot_path) && (quilt pop -a || true) && rm patches .pc -rf
```
ecourtois, line 21: I don't know what is quilt, but it has to be installed first, could you update the README?

patches/0001-sys-newlib_syscalls_default-set-_init-and-_fini-as-w.patch
```diff
@@ -0,0 +1,75 @@
...
+From: Gilles DOFFE <g.doffe@gmail.com>
+Date: Mon, 1 Aug 2022 15:35:24 +0200
+Subject: [PATCH] sys/newlib_syscalls_default: set _init() and _fini() as weak
+
+The _init() and _fini() functions are already defined into the
+C RunTime (crti.o) object file provided by the arm-none-eabi toolchain.
+To avoid multiple definition error, set them as 'weak'.
+
+Signed-off-by: Gilles DOFFE <g.doffe@gmail.com>
+---
```
ecourtois, line 47: The beginning of the patch is weird, the header appears 4 times...

ecourtois (changes requested):
There is a typo in the first commit message: `mae distclean`.

Maybe we could name the pathes directory `riot-patches` to make it more explicit.

Could create an issue related to this PR?

Usefull PR anyway!

patches/0001-sys-newlib_syscalls_default-set-_init-and-_fini-as-w.patch
gdoffe: :fearful:

Makefile
gdoffe: done

patches/0001-sys-newlib_syscalls_default-set-_init-and-_fini-as-w.patch
gdoffe: Dunno why :open_mouth: but fixed

ecourtois (approved):
LGTM!

PR #81, opened 2022-08-04
ztimer & cpp11-compat removal

gdoffe:
* replace xtimer by ztimer
* remove cpp11-compat calls and use low level RIOT API to prepare migration to libstdcpp


ecourtois (approved):
Tested with `uartpb` example on `cogip-nucleo-f446re` board.

LGTM!

PR #90, opened 2022-08-09
sys: shell_menu: remove dynamic allocation

ecourtois:
Refactor code to only use strings and containers allocated at compile time.

### DIFF
--- a/.github/workflows/compilation.yml
+++ b/.github/workflows/compilation.yml
@@ -31,7 +31,7 @@ jobs:
     - name: Install toolchain
       run: |
         sudo apt-get update
-        sudo apt-get install -qq build-essential gcc-multilib g++-multilib gcc-arm-none-eabi libnewlib-arm-none-eabi uncrustify protobuf-compiler
+        sudo apt-get install -qq build-essential gcc-multilib g++-multilib gcc-arm-none-eabi libnewlib-arm-none-eabi uncrustify protobuf-compiler quilt
         gcc --version
         arm-none-eabi-gcc --version
     - name: Install Python packages
--- a/examples/lidar_obstacles/lidar_obstacles.cpp
+++ b/examples/lidar_obstacles/lidar_obstacles.cpp
@@ -3,6 +3,7 @@
 
 // RIOT includes
 #include <ztimer.h>
+#include "time_units.h"
 
 // Project includes
 #include "lds01.h"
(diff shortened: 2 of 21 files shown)


examples/lidar_obstacles/lidar_obstacles.cpp
```diff
@@ -145,7 +145,7 @@ static void* _thread_obstacle_updater(void *data)
             _update_dynamic_obstacles_from_lidar(lidar_obstacles, *robot_state, filtered_distances);
         }
 
-        ztimer_periodic_wakeup(ZTIMER_USEC, &loop_start_time, TASK_PERIOD_USEC);
+        ztimer_periodic_wakeup(ZTIMER_MSEC, &loop_start_time, TASK_PERIOD_MSEC);
```
gdoffe, line 148: If you change it here, you have to change there also:
https://github.com/cogip/mcu-firmware/blob/6bbf79c40fbc383534e2aa9afa557d4ed30597d3/examples/lidar_obstacles/lidar_obstacles.cpp#L134

I was not sure of the accuracy of ZTIMER_MSEC, ztimer is quite complex and ZTIMER_MSEC could have different origin according to the architecture.
However I am quite confident on the accuracy now and we could apply this kind of fix to all the code. I will do it in a new PR.

.github/workflows/compilation.yml
```diff
@@ -31,7 +31,7 @@ jobs:
     - name: Install toolchain
       run: |
         sudo apt-get update
-        sudo apt-get install -qq build-essential gcc-multilib g++-multilib gcc-arm-none-eabi libnewlib-arm-none-eabi uncrustify protobuf-compiler
+        sudo apt-get install -qq build-essential gcc-multilib g++-multilib gcc-arm-none-eabi libnewlib-arm-none-eabi uncrustify protobuf-compiler quilt
```
gdoffe, line 34: s/quild/quilt/ in commt message

gdoffe (changes requested):

gdoffe (approved):
LGTM ! Good job !

examples/lidar_obstacles/lidar_obstacles.cpp
ecourtois: Ok I reverted this patch and included `time_units.h` to define `US_PER_MS`.
But I still do not understand why it is not defined in this file since it is in all other files without extra includes...

PR #91, opened 2022-08-09
boards: cogip-nucleo-f446re: split motor driver configuration

gdoffe:
This commit introduces no functional changes.
Other boards are using different pwm devices for each motor and
have one hardware driver per motor.
To reproduce this configuration for consistency with other boards and
allow this board to be used in applications, split the dual motor driver
into two single motor drivers.
PWM device is configured twice in the same way which is acceptable.

Signed-off-by: Gilles DOFFE <g.doffe@gmail.com>


ecourtois (commented):
LGTM!

ecourtois (approved):
LGTM!

PR #93, opened 2022-08-11
makefiles: fix RIOT patches application on specific application build

gdoffe:
If patches have never been applied and a specific application is built,
patches are not applied. For example:
  $ make -C applications/app_test

This commit fix this by adding patch target as a dependency for all apps
in Makefile.include.

Signed-off-by: Gilles DOFFE <g.doffe@gmail.com>


ecourtois (approved):
Sounds good!

PR #95, opened 2022-08-19
Use fixed size strings and containers, remove dynamic allocation

ecourtois:


gdoffe (changes requested):
I love this PR ! Good job @ecourtois ! So much important stuff !

Just an oversight I think, detail a bite more this commit, at least write a more explicit title:
https://github.com/cogip/mcu-firmware/pull/95/commits/7089fdbe8fc0c13b447bd4e30a5a5b9ef2d27c22

gdoffe (approved):

PR #112, opened 2022-09-27
makefiles: check glibc version before use of mallinfo2

gdoffe:
The mallinfo2 structure and function were introduced in glibc 2.33. Thus check glibc version on native architecture to avoid undefined reference to mallinfo2.


ecourtois (approved):
Thanks!

PR #113, opened 2022-10-02
Move planner from STM32 to Raspberry Pi

ecourtois:


lib/cogip_defs/include/cogip_defs/Pose.hpp
```diff
@@ -63,7 +63,7 @@ class Pose : public Coords {
         pose.set_O(O_);
     };
 
-private:
+protected:
```
gdoffe, line 66: Why this change ?

lib/path/Pose.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+    if (max_speed_angular_ == LOW_SPEED_ANGULAR) {
+        path_pose.set_max_speed_angular_enum(PB_SpeedEnum::LOW);
+    }
+    else if (max_speed_angular_ == MAX_SPEED_ANGULAR) {
+        path_pose.set_max_speed_angular_enum(PB_SpeedEnum::MAX);
+    }
+    else {
+        path_pose.set_max_speed_angular_enum(PB_SpeedEnum::NORMAL);
+    }
```
gdoffe, line 84: not a switch/case fan ? xD

lib/path/include/path/Pose.hpp
```diff
@@ -52,6 +54,16 @@ class Pose : public cogip_defs::Pose {
     /// Execute action callback if set.
     virtual void act() const { if (act_) act_(); }
 
+    /// Initialize the object from a protobuf message.
```
gdoffe, line 57: Protobuf

controllers/common_controller/include/ctrl.hpp
```diff
@@ -60,6 +60,12 @@
 /* Project includes */
 #include "odometry.hpp"
 
+#ifdef MODULE_UARTPB
+#include "uartpb/UartProtobuf.hpp"
+
+inline constexpr cogip::uartpb::uuid_t pose_reached_uuid = 2736246403;  ///< pose reached uuid for uartpb
+#endif
```
gdoffe, line 67: Is it an UUID per message or per module ?

gdoffe, line 67: `#endif // MODULE_UARTPB`

gdoffe (commented):
LGTM, more RAM, more flash available for new conrollers !
Just few comments/questions before approval.

lib/path/include/path/Pose.hpp
ecourtois: Done

controllers/common_controller/include/ctrl.hpp
ecourtois: Per message, and a module can handle multiple messages.

lib/cogip_defs/include/cogip_defs/Pose.hpp
ecourtois: Because it is used if path::Pose, but it is not in the related commit.
Fixed.

lib/path/Pose.cpp
ecourtois: You cannot use a `switch/case` on floating-point values...

controllers/common_controller/include/ctrl.hpp
ecourtois: Done

gdoffe: :+1:

gdoffe (approved):
LGTM!

PR #114, opened 2022-10-03
107 sysmon scheduling monitoring

gdoffe:
Introduce threads loops number and periods overshots monitoring.


sys/sysmon/ThreadStatus.cpp
```diff
@@ -12,6 +12,8 @@ namespace sysmon {
 
 ThreadStatus::ThreadStatus() : MemoryStatus() {
     pid_ = 0;
+    loops_ = 0;
+    overshots_ = 0;
 }
```
ecourtois, line 17: Note that you could just write in the header:
```cpp
ThreadStatus::ThreadStatus() : MemoryStatus(), pid_(0), loops_(0), overshots_(0) {};
```

sys/sysmon/include/sysmon/ThreadStatus.hpp
```diff
@@ -47,12 +47,24 @@ class ThreadStatus: public MemoryStatus {
         /// Constructor
         ThreadStatus();
 
+        /// Increment thread loops number
+        inline void inc_loops(void) { loops_++; };
+        /// Increment thread overshots number
+        inline void inc_overshots(void) { overshots_++; };
+        /// Get thread loops number
+        inline uint32_t loops() const { return loops_; };
```
ecourtois, line 55: Your are using either `void f(void)` or `void f()`. It means the same in C++.
I prefer `void f()` which is used in almost all C++ libraries.

sys/sysmon/include/sysmon/ThreadStatus.hpp
```diff
@@ -47,12 +47,24 @@ class ThreadStatus: public MemoryStatus {
         /// Constructor
         ThreadStatus();
 
+        /// Increment thread loops number
+        inline void inc_loops(void) { loops_++; };
```
ecourtois, line 51: Inline is implicit in C++ when the definition is present in the header file like in this case.
Moreover, inlining will never happen if the definition is in the source file.

sys/sysmon/include/sysmon/sysmon.hpp
```diff
@@ -36,6 +36,8 @@ void display_heap_status(void);
 void display_threads_status(void);
 /// Start system monitoring thread
 void sysmon_start(void);
+/// Update threads scheduling status
+void update_thread_sched_status(kernel_pid_t, bool);
```
ecourtois, line 40: Why unamed parameters? => no doxygen.

sys/sysmon/sysmon.cpp
```diff
@@ -186,7 +186,7 @@ static void *_thread_status_updater(void *data)
 {
     (void)data;
 
-    ztimer_now_t loop_start_time = ztimer_now(ZTIMER_USEC);
+    ztimer_now_t loop_start_time = ztimer_now(ZTIMER_SEC);
```
ecourtois, line 190: I thought we should not use `ZTIMER_SEC`.
Is it because the thread period is long enough here?

sys/thread/include/thread/thread.hpp
```diff
@@ -0,0 +1,38 @@
...
+// RIOT includes
+#include <ztimer.h>
+
+#pragma once
+
+namespace cogip {
+
+namespace thread {
+
+/// Display heap memory status
```
ecourtois, line 31: Bad copy-paste?

sys/thread/include/thread/thread.hpp
```diff
@@ -0,0 +1,38 @@
...
+void thread_ztimer_periodic_wakeup(ztimer_clock_t *clock, uint32_t *last_wakeup, uint32_t period);
```
ecourtois, line 32: Doxygen missing on parameters.

ecourtois (changes requested):
Almost good, just some comments about C++ language.

sys/sysmon/ThreadStatus.cpp
gdoffe: done

sys/sysmon/include/sysmon/ThreadStatus.hpp
gdoffe: Done but should have been seen in previous PR. Painful and ugly to change all now...

sys/sysmon/include/sysmon/ThreadStatus.hpp
gdoffe: done... painful too :)

sys/sysmon/include/sysmon/sysmon.hpp
gdoffe: done

sys/sysmon/sysmon.cpp
gdoffe: No, just because it is not a critical task needing high accuracy so it is acceptable here to use ZTIMER_SEC

sys/thread/include/thread/thread.hpp
gdoffe: done

sys/thread/include/thread/thread.hpp
gdoffe: done

ecourtois (approved):
Nice PR!

PR #117, opened 2022-10-09
makefiles: introduce Makefile.dep

gdoffe:
Feature, module and package depencies are listed in Makefile.include which is dirty and less readable.
Move them to Makefile.dep and include it in Makefile.include.


ecourtois (approved):
Just fix `depencies` in commit message and you can merge.

gdoffe:
> Just fix `depencies` in commit message and you can merge.

Thx done.
Just waiting CI now. :)

PR #146, opened 2024-01-30
boards: remove cogip-nucleo-f446re

### LINKED ISSUE #116: Clean cogip-nucleo-f446re
Remove dead configuration and rely more on RIOT nucleo-f446re

gdoffe:
Only COGIP board is now used and is suitable for testing too. Moreover MCU are now STM32G431RB.

Fix #116


ecourtois (approved):
LGTM!

One question: could come back to `cogip-board` instead of `cogip-board-ng` since there is no more legacy board?

gdoffe:
> One question: could come back to `cogip-board` instead of `cogip-board-ng` since there is no more legacy board?

Yes, will be done in new issue: https://github.com/cogip/mcu-firmware/issues/150

PR #154, opened 2024-02-05
Split configuration of PIDs

ecourtois:
Instead of configuring PIDs using only one message for all, use a single message of each PID.
This is required to reduce PID Protobuf messages to fit into a CAN frame.


gdoffe (approved):
LGTM !

PR #163, opened 2024-06-10
161 support anti blocking on motion control

gdoffe:


platforms/pf-robot-motion-control/motion_control.cpp
```diff
@@ -477,6 +476,26 @@ void pf_motor_drive(const cogip::cogip_defs::Polar &command)
     int16_t left_command = (int16_t) std::max(std::min(command.distance() - command.angle(), (double)(std::numeric_limits<int16_t>::max()) / 2),
                                                   (double)(std::numeric_limits<int16_t>::min()) / 2);
 
+    if (pf_motion_control_platform_engine.pose_reached() == cogip::motion_control::target_pose_status_t::blocked) {
```
mlecriva, line 479: Not mandatory for today in my mind but it can be good to add a method named `is_blocked` or renaming this method with something like `get_pos_status`. Its name will perhaps be more meaningful for this specific case.

platforms/pf-robot-motion-control/motion_control.cpp
```diff
@@ -363,8 +367,8 @@ void pf_handle_target_pose(cogip::canpb::ReadBuffer &buffer)
         }
 
         // Target speed
-        target_speed.set_distance((platform_max_speed_linear_mm_per_period * target_pose.max_speed_ratio_linear()));
-        target_speed.set_angle((platform_max_speed_angular_deg_per_period * target_pose.max_speed_ratio_angular()));
+        target_speed.set_distance((platform_max_speed_linear_mm_per_period * target_pose.max_speed_ratio_linear()) / 100);
+        target_speed.set_angle((platform_max_speed_angular_deg_per_period * target_pose.max_speed_ratio_angular()) / 100);
```
mlecriva, line 371: I'm not sure to understand the reason to divide by 100 here ? Is this a unit conversion?

lib/path/include/path/Pose.hpp
```diff
@@ -64,6 +64,18 @@ class Pose : public cogip_defs::Pose {
         PB_PathPose &path_pose    ///< [out] Protobuf message to fill
         ) const;
 
+    /// Override operator ==
+    bool operator==(const Pose& other) {
+        return (
+            this->x() == other.x() &&
```
mlecriva, line 74: The `==` comparison between doubles can cause problems in this case? 
Shouldn't we introduce a tolerance ratio here to ensure good double comparison?

platforms/pf-robot-motion-control/motion_control.cpp
```diff
@@ -434,6 +434,13 @@ void pf_motion_control_reset(void)
 void pf_disable_motion_control()
 {
     pf_motion_control_platform_engine.disable();
+
+    // Small wait to ensure engine is disabled
+    ztimer_sleep(ZTIMER_MSEC, motion_control_thread_period_ms);
```
mlecriva, line 438: Maybe it can be more safe to introduce an internal semaphore inside `enable/disable` methods instead of a timer ?

This is an area for improvement, probably not necessary now.

mlecriva (changes requested):

platforms/pf-robot-motion-control/motion_control.cpp
gdoffe: Yes due to legacy, it was just a boolean in the past. I think we will rework this in coming weeks with moving platform to c++.

lib/path/include/path/Pose.hpp
gdoffe: indeed, we could define a tolerance for float comparison like:

```C
fabs(this->x()-other.x()) < tolerance;
```

where `tolerance = 1.0e-2`;

platforms/pf-robot-motion-control/motion_control.cpp
gdoffe: true. i'll check this as we have time to do it.

platforms/pf-robot-motion-control/motion_control.cpp
```diff
@@ -126,7 +126,11 @@ static cogip::motion_control::PosePIDController linear_pose_controller(&linear_p
 static cogip::motion_control::SpeedFilterParameters linear_speed_filter_parameters(
     platform_min_speed_linear_mm_per_period,
     platform_max_speed_linear_mm_per_period,
-    platform_max_acc_linear_mm_per_period2
+    platform_max_acc_linear_mm_per_period2,
+    true,
```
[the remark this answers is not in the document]
gdoffe: done

lib/path/include/path/Pose.hpp
gdoffe: done

platforms/pf-robot-motion-control/motion_control.cpp
gdoffe: done

mlecriva (approved):
LGTM

PR #173, opened 2024-10-20
172 motion control rework odometry class to use new encoder interface

mlecriva:
Improve trigonometry library by replacing while loops by a periodic modulo in order to limit angles.

This use a little more resources to compute but avoid the usage of blocking loop inside limit functions.

Create a class that use encoder objects to compute odometry.

This library only support approximation by segment.

Add -ng suffix in order to keep last odometry lib until  the refactoring job is done.


lib/odometry-ng/include/odometer/OdometerDifferential.hpp
```diff
@@ -0,0 +1,93 @@
...
+        pos_.y = y;
+        pos_.O = DEG2RAD(O);
+    }
+    
+
+    /// @brief Set the default odometry position
+    ///
+    /// @note this function should be called to reset robot position and defined a new default one
+    ///
+    /// @param pos postion reference
```
gdoffe, line 39: s/pos/pose

lib/odometry-ng/include/odometer/OdometerDifferential.hpp
```diff
@@ -0,0 +1,93 @@
...
+    void set_position(const cogip::cogip_defs::Pose &pos) override
```
gdoffe, line 40: Should we move to CamelCase ? :thinking:

lib/odometry-ng/include/odometer/OdometerDifferential.hpp
```diff
@@ -0,0 +1,93 @@
...
+        pose.set_y(pos_.y);
+        pose.set_O(RAD2DEG(pos_.O));
+    }
+    
+    /// @brief Copy current velocities using cogip def format
+    /// @note Data units:
+    ///         - linear: mm/s
+    ///         - O: deg/s
+    /// @param vel cogip::cogip_defs::Polar current velocities
+    void copy_velocities(cogip::cogip_defs::Polar &vel) override
```
gdoffe, line 64: s/vel/velocity

A long time ago ^^' we made the choice to give explicit names, even if longer.

lib/odometry-ng/include/odometer/OdometerDifferential.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+    /// @brief Setup the encoders objects
+    /// @param left_encoder Left encoder
+    /// @param right_encoder Right encoder
+    void setup(cogip::encoder::EncoderInterface &left_encoder,
+               cogip::encoder::EncoderInterface &right_encoder)
+    {
+        left_encoder_  = &left_encoder;
+        right_encoder_ = &right_encoder;
+    }
```
gdoffe, line 78: I think it should be done in the constructor with left_encoder_ and right_encoder_  declared as const as they are immuable.

lib/odometry-ng/OdometerDifferential.cpp
```diff
@@ -0,0 +1,44 @@
...
+namespace odometer {
+
+int OdometerDifferential::update()
+{
+    if (!left_encoder_ && !right_encoder_) {
+        return -1;
+    }
+    
+    /// Compute encoders wheels left and right linear delta in mm
+    const double dL = left_encoder_->get_angle_and_reset() * parameters_.left_wheel_diameter_mm() * parameters_.left_polarity();
```
gdoffe, line 19: left_encoder_ and right encoder_ should be tested as they can be nullptr

lib/odometry-ng/include/odometer/OdometerDifferential.hpp
```diff
@@ -0,0 +1,93 @@
...
+namespace cogip {
+
+namespace odometer {
+
+class OdometerDifferential: public OdometerInterface {
+
+public:
+    /// @brief Construct a new differential odometer object
+    /// @param parameters Odometer parameters
+    explicit OdometerDifferential(OdometerDifferentialParameters &parameters) : parameters_(parameters), left_encoder_(NULL), right_encoder_(NULL) {}
```
gdoffe, line 18: s/NULL/nullptr

But i think encoders should be passed in constructor and be marked as const

lib/odometry-ng/include/odometer/OdometerDifferentialParameters.hpp
```diff
@@ -0,0 +1,68 @@
...
+class OdometerDifferentialParameters {
+public:
+    /// @brief Construct a new Odometry Params object
+    /// @param left_wheel_diameter left wheel diameter value (mm)
+    /// @param right_wheel_diameter right wheel diameter value (mm)
+    /// @param axle_track Axle track value (mm)
+    /// @param left_polarity left wheel polarity no unit, 1 or -1
+    /// @param right_polarity riht wheel polarity no unit, 1 or -1
+    OdometerDifferentialParameters(double left_wheel_diameter, 
+    double right_wheel_diameter, double axle_track, double left_polarity, double right_polarity) : left_wheel_diameter_(left_wheel_diameter), right_wheel_diameter_(right_wheel_diameter), axle_track_(axle_track), left_polarity_(left_polarity), right_polarity_(right_polarity) {};
```
gdoffe, line 16: what about `track_width` instead of `axle_track` ?

lib/odometry-ng/include/odometer/OdometerDifferentialParameters.hpp
```diff
@@ -0,0 +1,68 @@
...
+    /// @brief Get the right encoder polarity
+    /// @return double Encoder polarity. -1 or 1
+    double right_polarity() const { return right_polarity_; }
+
+private:
+    double left_wheel_diameter_;
+    double right_wheel_diameter_;
+    double axle_track_;
+    double left_polarity_;
+    double right_polarity_;
```
gdoffe, line 63: should be all const.

polarity could be a uint8_t

lib/odometry-ng/include/odometer/OdometerDifferential.hpp
```diff
@@ -0,0 +1,93 @@
...
+        pos_.y = pos.coords().y();
+        pos_.O = DEG2RAD(pos.O());
+    }
+    
+    /// @brief Copy current position using cogip def format
+    /// @note Data units:
+    ///         - x, y: mm
+    ///         - O: deg
+    /// @param pose cogip::cogip_defs::Pose current pose
+    void copy_position(cogip::cogip_defs::Pose &pose) override
```
gdoffe, line 52: call it `set_position` as its signature is different it works thanks to c/c++ function overloading

gdoffe (changes requested):
Good job ! Few things but the philosophy seems good to me.

lib/odometry-ng/Odometry.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+namespace odometry {
+
+int Odometry::update()
+{
+    if (!left_encoder_ && !right_encoder_) {
+        return -1;
+    }
+    
+    /* Compute encoders wheels left and right linear delta in mm */
```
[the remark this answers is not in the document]
mlecriva: Done

lib/odometry-ng/OdometerDifferential.cpp
mlecriva: Because we use reference, we can't have a null reference here. there is no need to check

lib/odometry-ng/include/odometer/OdometerDifferential.hpp
mlecriva: This is more a getter that copy the current position to the input pose reference. Maybe i can rename it `get_position` ?

lib/odometry-ng/include/odometer/OdometerDifferentialParameters.hpp
mlecriva: They cannot be const because I provide APIs in order to change it in run-time. This can be useful later when we want to run odometry calibration for example

I prefer to keep it double in order to avoid some cast during odometry computation.

lib/odometry-ng/include/odometer/OdometerDifferentialParameters.hpp
mlecriva: switched to `track_width` this is a better name

lib/odometry-ng/include/odometer/OdometerDifferential.hpp
mlecriva: The encoder cannot be const because there are called inside the update method that need to modify call members to compute the position and velocity.

But I now pass the constructor reference on the constructor to avoid pointers.

lib/odometry-ng/include/odometer/OdometerDifferential.hpp
mlecriva: done

lib/odometry-ng/include/odometer/OdometerDifferential.hpp
mlecriva: fixed.

lib/odometry-ng/include/odometer/OdometerDifferential.hpp
mlecriva: On the other class like in motion control, methods use snake case, I think we can keep it like this.

lib/odometry-ng/include/odometer/OdometerDifferential.hpp
mlecriva: Fixed

lib/odometry-ng/Odometry.cpp
```diff
@@ -0,0 +1,44 @@
+#include <math.h>
+#include <stdlib.h>
```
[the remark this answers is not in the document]
mlecriva: Removed because it is unused.

lib/odometry-ng/Odometry.cpp
```diff
@@ -0,0 +1,44 @@
+#include <math.h>
```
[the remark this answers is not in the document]
mlecriva: Done.

PR #184, opened 2025-02-08
motion control: fix PassthroughPosePIDController parameters number

mlecriva:
The DualPIDMetaController took 5 parameter, and the SpeedFilter took 5 inputs parameters. So, the PassthroughPosePIDController should input and output the same number of parameters in order to be inserted in the DualPID meta controller.

### DIFF
--- a/motion_control/controllers/passthrough_pose_pid_controller/PassthroughPosePIDController.cpp
+++ b/motion_control/controllers/passthrough_pose_pid_controller/PassthroughPosePIDController.cpp
@@ -41,6 +41,8 @@ void PassthroughPosePIDController::execute() {
     this->outputs_[2] = parameters_->target_speed();
     // Store disabling speed filter (pass through)
     this->outputs_[3] = this->inputs_[3];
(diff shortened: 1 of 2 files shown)


motion_control/controllers/passthrough_pose_pid_controller/PassthroughPosePIDController.cpp
```diff
@@ -41,6 +41,8 @@ void PassthroughPosePIDController::execute() {
     this->outputs_[2] = parameters_->target_speed();
```
gdoffe, line 41: Fixed typos:
```
The DualPIDMetaController took 5 parameters, and the SpeedFilter took 5
input parameters. So, the PassthroughPosePIDController should input and
output the same number of parameters in order to be inserted in the
DualPID meta controller.
```

gdoffe (approved):
Small typo, fix it (copy/paste commit message :wink: ) , resolve the thread and merge !

Good catch !

PR #185, opened 2025-02-11
155 clean unused firmware components

gdoffe:
This PR removes from firmware all legacy code and unused modules.
Each deprecated module is removed in a specific commit in case of revert needed.


mlecriva (changes requested):
Seems good to me, only one comment about cogip-board-pami

boards/cogip-board/Makefile.include
```diff
@@ -6,3 +6,6 @@ PORT = "/dev/ttyUSB0"
 
 # Setup of programmer and serial is shared between STM32 based boards
 include $(RIOTMAKE)/boards/stm32.inc.mk
+
+# Set onboard transceiver loop delay
+CFLAGS += -DCONFIG_FDCAN_DEVICE_TRANSCEIVER_LOOP_DELAY=130
```
mlecriva, line 11: How do you determine this value ?

boards/cogip-board-pami/board.c
```diff
@@ -19,6 +19,7 @@
  */
```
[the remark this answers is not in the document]
gdoffe: Good catch, done.

boards/cogip-board/Makefile.include
gdoffe: You can find it inside the [datasheet](https://www.ti.com/lit/ds/symlink/tcan332g.pdf?ts=1739357704562&ref_url=https%253A%252F%252Fwww.mouser.fr%252F):
image

I choose a value that is inside both ranges, nominal and highly loaded network. i added this to the commit message:
```
The TCAN332GDCNR CAN transceiver datasheet exposes two ranges for this          
value:                                                                          
- Nominal range: 100 (typical) to 135 (max)                                     
- Highly loaded network: 120 (typical) to 180 (max)                             
                                                                                
So here the value 130 is choosen as it is inside both ranges.                   
This value has been tested in nominal robot working conditions.    
```

mlecriva: Ok ! Thank's

mlecriva (approved):

mlecriva:
LGTM

Issue #182, opened 2025-02-05
Motion Control - Fix wrong unit type for motion control differential drive


mlecriva:
As the platform engine is working using degrees. 

We should convert this degrees into radians to correctly convert the Polar speed into motor speeds.

PR #186, opened 2025-02-13
181 add new generic motion control application

mlecriva:
As we introduced Odometer and differential drive inside motion control
platform engine, we also need to update the platform accordingly.

This PR makes the platform works as same as before but with using
these new classes and the new pose_reached callback to deal with motion
control states.


lib/motor/include/motor/MotorDriverRIOT.hpp
```diff
@@ -55,27 +55,27 @@ class MotorDriverRIOT: public MotorDriverInterface {
     /// @param speed speed in % [-100; 100]
     /// @param id id of the motor
     /// @return  0 on success, negative on error
-    int speed(int speed, int id) override
+    int speed(double speed, int id) override
```
gdoffe, line 58: should be called `set_speed`

lib/motor/include/motor/MotorInterface.hpp
```diff
@@ -22,8 +22,8 @@ class MotorInterface {
     /// @brief Set motor speed
     /// @param speed speed in % [-100; 100]
     /// @return  0 on success, negative on error
-    virtual int speed(int speed) = 0;
-    
+    virtual int speed(double speed) = 0;
```
gdoffe, line 25: Same here: should be called set_speed

lib/motor/include/motor/MotorRIOT.hpp
```diff
@@ -30,11 +30,11 @@ class MotorRIOT: public MotorInterface {
     /// @brief Set motor speed
     /// @param speed speed in % [-100; 100]
     /// @return  0 on success, negative on error
-    int speed(int speed)
+    int speed(double speed)
```
gdoffe, line 33: Same here: should be called set_speed

motion_control/engines/platform_engine/Makefile.dep
```diff
@@ -1,4 +1,5 @@
 USEMODULE += motion_control_common
```
gdoffe, line 1: In `platform engine: replace callbacks by object references`, the scope is not good, should be:
`motion_control/engines: replace callbacks by object references`

lib/odometry-ng/OdometerDifferential.cpp
```diff
@@ -31,7 +31,7 @@ int OdometerDifferential::update()
     
     // Save polar pose delta since last call
     polar_.set_distance(delta_linear_pose);
-    polar_.set_angle(delta_angular_pose);
+    polar_.set_angle(RAD2DEG(delta_angular_pose));
```
gdoffe, line 34: I think that degrees should come to and from the RPi, but inside it should be radians.

gdoffe (changes requested):
Good job but some changes needed to avoid `#ifdef`

lib/odometry-ng/OdometerDifferential.cpp
mlecriva: I agree with you but I created a special issue for this https://github.com/cogip/mcu-firmware/issues/182 because this requires some additional changes that is out of this PR scope in my opinion.

platforms/pf-robot-motion-control/motion_control.cpp
```diff
@@ -174,6 +173,52 @@ static cogip::motion_control::PassthroughPosePIDController passthrough_angular_p
 /// Quad PID meta controller.
 static cogip::motion_control::QuadPIDMetaController quadpid_meta_controller;
 
+/// Encoders
+#ifdef CPU_NATIVE
```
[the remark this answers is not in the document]
mlecriva: I agree with you, but even if i switched to a classic enum, a `static_cast` is required.

`ENCODER_MODE_X?` was created in order to define a generic interface.

So, I chose to keep it and use a static cast around `QDEC_MODE`, removing the `#ifdef <SOME_ARCH>` usage and keep the following inside `platforms/pf-robot-motion-control/include/motion_motors_params.hpp`
```c++
/// Quadrature decoding
#ifndef QDEC_MODE
#define QDEC_MODE QDEC_X4
#endif
```

Another solution is to replace the whole enum by a simple `int` but I think we loose useful information in this case.

platforms/pf-robot-motion-control/motion_control.cpp
```diff
@@ -321,6 +350,201 @@ void pf_send_pid(PB_PidEnum id)
...
+    if (pf_motion_control_platform_engine.pose_reached() != cogip::motion_control::target_pose_status_t::reached) {
+        qdecs_value[MOTOR_RIGHT] = (linear_speed_filter.previous_speed_order() * pulse_per_mm +
+                        angular_speed_filter.previous_speed_order() * pulse_per_degree / 2) *
+                       QDEC_RIGHT_POLARITY;
+        qdecs_value[MOTOR_LEFT] = (linear_speed_filter.previous_speed_order() * pulse_per_mm -
+                       angular_speed_filter.previous_speed_order() * pulse_per_degree / 2) *
+                      QDEC_LEFT_POLARITY;
+    }
+}
+#endif
```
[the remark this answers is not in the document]
mlecriva: I agree with you, but this is more complicated for this because `cogip_native_motor_driver_qdec_simulation` uses some objects defined inside the platform.

Moving it in `boards/cogip-native/board.c` requires to link the platform inside the board and I think this is not a good idea.

Maybe we can find another solution but this is legacy from the previous platform and maybe it's better to keep it like this for now. What do you think ?

lib/odometry-ng/OdometerDifferential.cpp
gdoffe: ok good to me. i resolve.

gdoffe (approved):
LGTM ! :partying_face:

PR #188, opened 2025-02-26
187 change all doubles for floats

gdoffe:
While the RAM savings are minimal, using float instead of double aligns
better with the STM32G474RE’s FPU capabilities, avoiding unnecessary
software emulation and improving computation efficiency.


mlecriva (approved):
LGTM

PR #194, opened 2025-05-02
190 update lift motor firmware for cup 2025

gdoffe:
* create an actuator library module
* clean pf-robot-motors from application specific code
* update the application accordingly and rename cup2024-robot-motors to cup2025-robot1-lift
* remove unused pf-robot-actuators platform and cup2024-robot-actuators application


lib/motor/include/motor/MotorRIOT.hpp
```diff
@@ -13,6 +13,13 @@ class MotorRIOT: public MotorInterface {
     {
     }
 
+    /// @brief Init the motor and its driver
+    /// @return 0 on success, negative on error
+    int init()
+    {
+        return driver_.init();
```
mlecriva, line 20: Do you check if we can safely call driver init multiple times for the same driver ?

platforms/pf-robot-motion-control/motion_control.cpp
```diff
@@ -674,12 +674,12 @@ void pf_init_motion_control(void)
     // Init motor driver
     motor_driver.init();
```
mlecriva, line 676: You can remove this line, as you moved the initialisation set inside the motor implementation.

lib/actuator/LiftsLimitSwitchesManager.cpp
```diff
@@ -0,0 +1,99 @@
...
+        event_stack_, sizeof(event_stack_),
+        THREAD_PRIORITY_MAIN - 1,
+        THREAD_CREATE_STACKTEST,
+        event_thread_entry, nullptr, "gpio_evt"
+    );
+
+    std::cout << "Motor limit switches manager started" << std::endl;
+}
+
+void LiftsLimitSwitchesManager::register_gpio(gpio_t pin, cogip::actuators::positional_actuators::Lift* lift) {
```
mlecriva, line 26: You can return appropriates error codes here, it can be useful for error management.

lib/actuator/Motor.cpp
```diff
@@ -0,0 +1,111 @@
...
+      speed_controller_(motor_parameters.speed_controller_parameters),
+      motor_distance_filter_(motor_parameters.motor_pose_filter_parameters),
+      speed_filter_(motor_parameters.speed_filter_parameters),
+      motor_engine_(motor_parameters.motor, motor_parameters.odometer, motor_parameters.engine_thread_period_ms)
+{
+    // Build the cascade: MotorPoseFilter → PosePID → SpeedFilter → SpeedPID
+    dualpid_meta_controller_.add_controller(&motor_distance_filter_);
+    dualpid_meta_controller_.add_controller(&distance_controller_);
+    dualpid_meta_controller_.add_controller(&speed_filter_);
+    dualpid_meta_controller_.add_controller(&speed_controller_);
```
mlecriva, line 35: Interesting way to use the engines. We can think about use this in the same way for the Motion control platform ?

lib/actuator/Motor.cpp
```diff
@@ -0,0 +1,111 @@
...
+        std::cerr << "Odometer init failed" << std::endl;
+    }
+
+    // Ensure overload flag is cleared at startup
+    gpio_init(params_.clear_overload_pin, GPIO_OUT);
+    gpio_clear(params_.clear_overload_pin);
+
+    // Start disabled and spin up the control thread
+    disable();
+    motor_engine_.disable();
```
mlecriva, line 56: This line can be removed as you call it inside `disable` method.

lib/actuator/include/actuator/LiftsLimitSwitchesManager.hpp
```diff
@@ -0,0 +1,109 @@
...
+namespace positional_actuators {
+
+/// @brief Singleton managing GPIO-based limit switches for motors.
+/// @details This class handles GPIO interrupts via a dedicated thread,
+///          allowing motor control to react to events such as limit switches
+///          in a safe, non-blocking, and asynchronous way.
+class LiftsLimitSwitchesManager {
+public:
+    /// @brief Callback type used to react to GPIO events in thread context.
+    using GpioCallback = etl::delegate<void(void)>;
```
mlecriva, line 38: I didn't see any uses of `GpioCallback` ? Maybe I miss something or this line should be removed.

lib/actuator/include/actuator/LiftsLimitSwitchesManager.hpp
```diff
@@ -0,0 +1,109 @@
...
+    static void isr_callback(void* arg);
+
+    /// @brief Entry function for the event processing thread.
+    /// @param arg Unused thread argument.
+    /// @return nullptr on exit.
+    static void* event_thread_entry(void* arg);
+
+    /// @brief Dispatches a GPIO event to the registered Lift callback.
+    /// @param pin GPIO pin whose event occurred.
+    void handle_event(gpio_t pin);
```
mlecriva, line 74: Maybe it can be renamed to `handle_gpio_event` to avoid confusing with `event_handler` as their names are close.

lib/actuator/include/actuator/MotorParameters.hpp
```diff
@@ -0,0 +1,75 @@
...
+    cogip::motion_control::PosePIDControllerParameters* pose_controller_parameters = nullptr;
+
+    /// @brief Parameters for the SpeedPIDController (speed control).
+    cogip::motion_control::SpeedPIDControllerParameters* speed_controller_parameters = nullptr;
+
+    /// @brief Parameters for the MotorPoseFilter (used to smooth and validate position).
+    cogip::motion_control::MotorPoseFilterParameters* motor_pose_filter_parameters = nullptr;
+
+    /// @brief Parameters for the SpeedFilter (used to smooth and limit speed).
+    cogip::motion_control::SpeedFilterParameters* speed_filter_parameters = nullptr;
```
mlecriva, line 59: I think the whole controllers parameters pointers can be replaced by references ? As we always need to define it.

applications/cup2024-robot-motors/include/real/app_conf.hpp
```diff
@@ -20,3 +55,182 @@ constexpr float motor_lift_speed_pid_integral_limit = 500 / motor_lift_speed_pid
...
+    /* reverse_polarity */ true
+};
+
+/// @brief Lift motor odometer
+static cogip::localization::OdometerEncoder lift_motor_odometer(lift_motor_odometer_params, lift_motor_encoder);
+
+// 1) Prepare generic motor parameters for the front lift actuator
+static const cogip::actuators::positional_actuators::MotorParameters lift_motor_params{
+    /* id                           */ cogip::actuators::Enum::MOTOR_LIFT,
+    /* default_timeout_ms           */ 2000u,
```
mlecriva: This can be declared as a constant with explicit name. Same for other hardcoded values in this file

applications/cup2024-robot-motors/include/real/app_conf.hpp
```diff
@@ -20,3 +55,182 @@ constexpr float motor_lift_speed_pid_integral_limit = 500 / motor_lift_speed_pid
...
+
+// 2) Specify front-lift-specific speeds (up/down)
+static const cogip::actuators::positional_actuators::LiftParameters lift_params{
+    /* motor_params             */ lift_motor_params,       ///< Base motor parameters
+    /* init_speed_percentage    */ 100
+                                   * lift_limits::max_init_speed_mm_per_period
+                                   / lift_limits::max_speed_mm_per_period,
+                                                            ///< Init sequence speed
+    /* lower_limit              */ 0,                       ///< Lower lift limit
+    /* upper_limit              */ 160,                     ///< Upper lift limit
```
mlecriva: You can precise the units for both lower and upper limits

mlecriva (changes requested):
Good job ! It's nice to see the codebase moving to the object model

lib/motor/include/motor/MotorRIOT.hpp
gdoffe: yes, in case of double h-bridge, the driver uses the same PWM frequency/resolution for both motors:
https://github.com/RIOT-OS/RIOT/blob/bf8c734f988c7f6686e23b3b28aee9aae9658da1/drivers/include/motor_driver.h#L168

Here we have 2 simple h-bridge, but seen as one double because we apply the same configuration.

So yes it is completly safe. Moreover if a mistake has to be done, it will be in the motor param file, so the risk is the same than doing the init outside of the class.

platforms/pf-robot-motion-control/motion_control.cpp
gdoffe: Done.

lib/actuator/Motor.cpp
gdoffe: The `add_controler` method ? We are already using it:
https://github.com/cogip/mcu-firmware/blob/c3b80e1ecb8c940d0485bf058b187f6ad056db3a/platforms/pf-robot-motion-control/motion_control.cpp#L221

lib/actuator/Motor.cpp
gdoffe: Done

lib/actuator/include/actuator/LiftsLimitSwitchesManager.hpp
gdoffe: Done.

lib/actuator/include/actuator/LiftsLimitSwitchesManager.hpp
gdoffe: Oooops oversight ! Done.

lib/actuator/include/actuator/MotorParameters.hpp
gdoffe: Done. I think it could also be replaced by references inside the motion_control library but I prefer to keep that modification for motion control incoming rework.

applications/cup2024-robot-motors/include/real/app_conf.hpp
gdoffe: Done.

applications/cup2024-robot-motors/include/real/app_conf.hpp
gdoffe: Done.

lib/actuator/LiftsLimitSwitchesManager.cpp
gdoffe: Done

lib/actuator/Motor.cpp
mlecriva: Yes but it's more about wrapping it inside a class

mlecriva (approved):
LGTM good job !

PR #208, opened 2025-10-05
207 allow to run controllers at various frequency

gdoffe:
## Summary

  Implement proper cascaded PID control with different frequencies for pose and speed loops.

  - Add `ThrottledController` wrapper class to execute controllers at configurable frequencies
  - Split control architecture into pose loop (200ms) and speed loop (20ms)
  - Pose loop (outer loop) runs at 5 Hz for trajectory tracking
  - Speed loop (inner loop) runs at 50 Hz for fast motor response
  - Adjust robot1 angular threshold for the new control loop timing

  ## Architecture

  Engine (20ms) → QuadPID:
  1. ThrottledController (÷10) → Pose Loop (200ms)
    - pose_straight_filter
    - linear/angular pose PIDs
  2. Speed Loop (20ms)
    - linear/angular speed filters
    - linear/angular speed PIDs

  ## Rationale

  Cascaded PID control requires the inner loop (speed) to run faster than the outer loop (pose) to ensure stability and proper control
  hierarchy. This implementation follows standard control theory practice where the inner loop should be 5-10x faster than the outer
  loop.

### DIFF
--- a/applications/robot-motion-control/include/robot1_conf.hpp
+++ b/applications/robot-motion-control/include/robot1_conf.hpp
@@ -46,7 +46,7 @@ constexpr float angular_speed_pid_kd = 0;
 // Linear threshold
 constexpr float linear_threshold = 1;
 // Angular threshold
-constexpr float angular_threshold = 1;
+constexpr float angular_threshold = 2;
 // Angular intermediate threshold (when the robot turns on itself to go straight
 // to its destination)
 constexpr float angular_intermediate_threshold = 20;


motion_control/motion_control_common/ThrottledController.cpp
```diff
@@ -0,0 +1,55 @@
...
+#define ENABLE_DEBUG 0
+#include <debug.h>
+#include <inttypes.h>
+
+namespace cogip {
+
+namespace motion_control {
+
+ThrottledController::ThrottledController(BaseController* wrapped_controller,
+                                         uint32_t period_divider)
```
mlecriva, line 17: Maybe you can use a smaller type. `uint16_t` is enough I think for all kind of use cases. But this is a matter of optimisation, so feel free to ignore this comment. (Maybe a `uint8_t` can also do the job)

motion_control/motion_control_common/ThrottledController.cpp
```diff
@@ -0,0 +1,55 @@
...
+    : BaseController(), wrapped_controller_(wrapped_controller), period_divider_(period_divider),
+      current_count_(0)
+{
+    // Ensure period_divider is at least 1
+    if (period_divider_ < 1) {
+        period_divider_ = 1;
```
mlecriva, line 23: Maybe you can add a warning here. This can be useful for development purposes

applications/robot-motion-control/include/robot1_conf.hpp
```diff
@@ -46,7 +46,7 @@ constexpr float angular_speed_pid_kd = 0;
 // Linear threshold
 constexpr float linear_threshold = 1;
 // Angular threshold
-constexpr float angular_threshold = 1;
+constexpr float angular_threshold = 2;
```
mlecriva, line 49: Maybe you can verify this point during real testing session.

I think we should be able to maintain control loop accurate as before.

Maybe try this two configuration and check the impact in robot stability:
- motion control engine (Odometry) @100Hz -> postion loop @20Hz -> Speed loop @100Hz (~5 ratio)
- motion control engine (Odometry) @200Hz-> postion loop @20Hz -> Speed loop @200Hz (~10 ratio)

The goal is to keep position control loop as faster as before. My assumption is that 100ms or 200ms is too high to preserve robot accuracy and faster response time.

mlecriva (changes requested):
Good Job ! Few remarks

PR #209, opened 2025-10-08
platforms: centralize common platform logic in pf-common module

### LINKED ISSUE #204: Centralize CAN message UUIDs in shared header file
The CAN message UUIDs and their range definitions are currently duplicated across platform-specific files (e.g., platforms/pf-power-supply/include/platform.hpp). 
This creates a risk of inconsistencies and makes maintenance difficult.

Move the following to a common shared header:
  - Message ID range definitions (Motion Control, Actuators, Service, Game, Power Supply, Board)
  - All cogip::canpb::uuid_t constants

This ensures all applications share the same UUID definitions and prevents conflicts.

mlecriva:
Create pf-common platform module to consolidate duplicate initialization code across platform implementations.

Changes:
- Add pf-common module with CAN, heartbeat, and copilot connect/disconnect handlers
- Centralize CAN message UUIDs in pf-common/uuids.hpp
- Refactor pf-robot-motion-control to use pf-common base
- Refactor pf-robot-motors to use pf-common base
- Refactor pf-power-supply to use pf-common base with custom callback for on-connect status send

Benefits:
- Single source of truth for CAN message UUIDs
- Reduced code duplication
- Easier maintenance and consistency
- Proper separation of common vs platform-specific logic

closes #204


platforms/pf-robot-motion-control/include/platform.hpp
```diff
@@ -22,34 +22,33 @@
...
+using cogip::pf_common::blocked_uuid;
+using cogip::pf_common::brake_uuid;
+using cogip::pf_common::controller_uuid;
+using cogip::pf_common::intermediate_pose_reached_uuid;
+using cogip::pf_common::pid_request_uuid;
+using cogip::pf_common::pid_uuid;
+using cogip::pf_common::pose_order_uuid;
+using cogip::pf_common::pose_reached_uuid;
+using cogip::pf_common::pose_start_uuid;
+using cogip::pf_common::state_uuid;
```
gdoffe: really useful here ? Should it be inside `pf_common/uuids.hpp` before namespace declaration ?

gdoffe (changes requested):

platforms/pf-robot-motion-control/include/platform.hpp
mlecriva: The goal is to explicitly "include" these UUIDs as part of the motion control platform so that they can be used directly within it, like this: `game_reset_uuid`; instead of using `cogip::pf_common::game_reset_uuid;`

This is the way I choose to explicitly declare that certain UUIDs from `pf_common/uuids.hpp` are part of and can be used within this platform.

gdoffe: So some kind of "I allow (somehow) this code to use these UUIDs"

gdoffe (approved):
LGTM, lets go merge !

PR #225, opened 2025-12-11
189 add odometry tuning support

mlecriva:
# Summary
  - Add accumulated tick counter to encoder interface for tracking total encoder position
  - Implement generic thread-safe telemetry library for streaming data over CAN protocol buffer
  - Move FNV-1a key hash to utils library for reuse across telemetry and parameters
  - Integrate encoder telemetry in pf-robot-motion-control platform for odometry tuning

### DIFF
--- a/platforms/pf-robot-motion-control/motion_control.cpp
+++ b/platforms/pf-robot-motion-control/motion_control.cpp
@@ -44,12 +44,15 @@
 #include "PB_Pid.hpp"
 #include "PB_State.hpp"
 #include "etl/limits.h"
+#include "telemetry/Telemetry.hpp"
 
 namespace cogip {
 
 namespace pf {
 
 namespace motion_control {
+
+using cogip::utils::operator"" _key_hash;
 
 // Current controller
 static uint32_t current_controller_id = 0;
@@ -621,6 +624,12 @@ void pf_send_pb_state(void)
     pf_motion_control_platform_engine.target_speed().pb_copy(pb_state.mutable_speed_order());
 
     pf_get_canpb().send_message(state_uuid, &pb_state);
+}
+
+void pf_send_encoder_telemetry(void)
+{
+    cogip::telemetry::Telemetry::send<int64_t>("encoder_left"_key_hash, left_encoder.counter());
+    cogip::telemetry::Telemetry::send<int64_t>("encoder_right"_key_hash, right_encoder.counter());
 }
 
 void pf_handle_brake([[maybe_unused]] cogip::canpb::ReadBuffer& buffer)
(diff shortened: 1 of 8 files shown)


platforms/pf-robot-motion-control/motion_control.cpp
```diff
@@ -623,6 +629,24 @@ void pf_send_pb_state(void)
     pf_get_canpb().send_message(state_uuid, &pb_state);
 }
 
+void pf_send_pb_telemetry(void)
```
gdoffe: Why not passing telemetry type as argument ? 🤔
Because it seems quite complex to first set telemetry type and then send it in an other function. Race condition could occur in the future with other type of telemetry. 
I think protecting the function with a mutex could also be useful. What do you think ?

gdoffe (changes requested):

lib/encoder/include/encoder/EncoderInterface.hpp
```diff
@@ -70,6 +70,16 @@ class EncoderInterface
     ///
     virtual void reset() = 0;
 
+    ///
+    /// @brief Get the accumulated encoder counter value.
+    ///
+    /// @return int64_t accumulated counter value in ticks.
+    ///
+    int64_t get_counter()
```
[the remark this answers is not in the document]
mlecriva: Done.

examples/parameter_demo/main.cpp
```diff
@@ -11,14 +11,15 @@
 
 #include "log.h"
```
gdoffe: in commit ` lib: utils: move FNV-1a key hash from parameter to utils ` message, you explain the what but not the why of this commit. A sentence is enough, but thats the reason of the commit that matters.

lib/telemetry/include/telemetry/Telemetry.hpp
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        PB_TelemetryData message;
+        message.set_key_hash(key_hash);
+        message.set_timestamp_ms(ztimer_now(ZTIMER_MSEC));
+
+        if constexpr (etl::is_same<T, float>::value) {
+            message.set_float_value(value);
+        } else if constexpr (etl::is_same<T, double>::value) {
+            message.set_double_value(value);
```
gdoffe: I would drop double type.

lib/telemetry/PB_Telemetry.proto
```diff
@@ -0,0 +1,23 @@
...
+
+syntax = "proto3";
+
+message PB_TelemetryData {
+    uint32 key_hash = 1;     ///< FNV-1a hash of the telemetry key
+    uint32 timestamp_ms = 2; ///< Timestamp in milliseconds
+
+    oneof value {
+        float float_value = 3;   ///< 32-bit float
+        double double_value = 4; ///< 64-bit double
```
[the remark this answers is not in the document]
mlecriva: removed.

examples/parameter_demo/main.cpp
mlecriva: Every commits has been reworded.

platforms/pf-robot-motion-control/motion_control.cpp
gdoffe: Refactored, we are good here.

PR #228, opened 2025-12-19
motion_control: fix "error getting pose_reached" by initializing IO and improving enum handling

gdoffe:
- Initialize pose_reached to moving in PlatformEngine::prepare_inputs() so the key exists before process_outputs() reads it
- Store enums as int in ControllersIO variant instead of target_pose_status_t directly
- Update get_as<>() to retrieve enums by fetching int and casting to the requested enum type
- Increase MAX_PARAMS from 32 to 48


mlecriva (approved):
LGTM

PR #233, opened 2026-02-16
231 update lift code with new motion control and multiple lifts

### LINKED ISSUE #231: Update lift code with new motion control and multiple lifts
Motion control stack got a lot of changes recently. Even if lift app is buildable, it does not work as is.
Moreover, canpb uuid are for one lift, not several lifts. A way must be find to manage several lifts.

gdoffe:
Summary                                         
                                                                                                                                                                                                                                                                                                                              - Add DUALPID_TRACKER control mode to Motor using a trapezoidal velocity profile with tracking error correction, replacing the reactive cascaded PID for lifts
- Replace cup2025-robot1-lift with a generic robot-lift-control application supporting build-time LIFT_ID selection (1 or 2), each with its own config (mechanical limits, PID gains, tracker parameters)                                                                                                                   
- Make ParameterHandler return optional responses so only the board that owns a parameter replies on the shared CAN bus (multi-board support)
- Clear motor driver overload faults continuously from the control loop instead of only on new commands


motion_control/motion_control_common/BaseControllerEngine.cpp
```diff
@@ -76,6 +76,10 @@ void BaseControllerEngine::thread_loop()
                 LOG_ERROR("Engine timed out\n");
 
                 enable_ = false;
+
+                // Zero out speed orders before processing outputs
```
mlecriva: We merge this commit in last MR, is this the same case covered ?
https://github.com/cogip/mcu-firmware/commit/8acec086a2c10ac619a3c5850e799a6a7ad5290f

Maybe we would like to avoid dealing with io in base controller, WDYT?

lib/actuator/Lift.cpp
```diff
@@ -33,29 +41,41 @@ void Lift::init()
...
-                  static_cast<uint8_t>(id_));
-    }
-    ret = LiftsLimitSwitchesManager::instance().unregister_gpio(params_.upper_limit_switch_pin);
-    if (ret) {
-        LOG_ERROR("Upper switch pin for lift %" PRIu8 " unregistering failed\n",
-                  static_cast<uint8_t>(id_));
-    }
+    LOG_INFO("Lift init complete\n");
+
+    // Keep limit switches registered for runtime protection
```
mlecriva: This comment can be removed

lib/actuator/LiftsLimitSwitchesManager.cpp
```diff
@@ -3,7 +3,7 @@
 #include <inttypes.h>
 #include <thread.h>
 
-#define ENABLE_DEBUG 0
+#define ENABLE_DEBUG 1
```
mlecriva: You probably forgot to keep it back to 0

lib/actuator/Motor.cpp
```diff
@@ -4,44 +4,106 @@
...
+        // Build tracker chain:
+        // MotorPoseFilter → ProfileTracker → PosePID → Combiner →
+        //   [SpeedLimitFilter] → [AccelerationFilter] → SpeedPID
+        tracker_meta_controller_.add_controller(&tracker_motor_distance_filter_);
+        tracker_meta_controller_.add_controller(&profile_tracker_controller_);
+        tracker_meta_controller_.add_controller(&tracker_pose_controller_);
+        tracker_meta_controller_.add_controller(&tracker_combiner_controller_);
+        // Safety filters (if configured) - applied to speed_order before SpeedPID
+        if (motor_parameters.deceleration_filter_parameters) {
+            tracker_deceleration_filter_ = new motion_control::DecelerationFilter(
```
mlecriva: This is dynamic memory allocation, we probably want to avoid it.

Maybe you can resolve this statically, WDYT ? or maybe with some etl::optional ?

mlecriva: Available for the whole file

platforms/pf-robot-motors/pf_actuators.cpp
```diff
@@ -18,6 +18,9 @@
 
 #include "PB_Actuators.hpp"
 
+#define ENABLE_DEBUG 1
```
mlecriva: It's maybe better to set it to 0 here

platforms/pf-robot-motors/pf_positional_actuators.cpp
```diff
@@ -19,6 +19,9 @@
 #include "etl/pool.h"
 #include "log.h"
 
+#define ENABLE_DEBUG 1
```
mlecriva: Same here

lib/actuator/Motor.cpp
```diff
@@ -89,6 +89,12 @@ Motor::Motor(const MotorParameters& motor_parameters, MotorControlMode mode)
             tracker_meta_controller_.add_controller(tracker_acceleration_filter_);
         }
         tracker_meta_controller_.add_controller(&tracker_speed_controller_);
+        if (motor_parameters.anti_blocking_parameters) {
+            anti_blocking_controller_ = new motion_control::AntiBlockingController(
```
mlecriva: Same here, we should avoid dynamic allocation

mlecriva (changes requested):
Nice idea about multi-board parameters support.

Good job, a few remarks about dynamic allocation

motion_control/motion_control_common/BaseControllerEngine.cpp
gdoffe: Agreed. Already handled by commit 8acec08 in PlatformEngine. Removed this commit entirely as it was an oversight. I already removed a lot of outdated commits.

lib/actuator/Lift.cpp
gdoffe: Done, removed.

lib/actuator/Motor.cpp
gdoffe: Replaced all new with etl::optional<T> and .emplace(). No more dynamic allocation.

platforms/pf-robot-motors/pf_positional_actuators.cpp
gdoffe: Done, set to 0.

platforms/pf-robot-motors/pf_actuators.cpp
gdoffe: Done, set to 0.

lib/actuator/Motor.cpp
gdoffe: Replaced all new with etl::optional<T> and .emplace(). No more dynamic allocation.

lib/actuator/LiftsLimitSwitchesManager.cpp
gdoffe: Done, set to 0.

mlecriva (approved):
LGTM nice improvement !

PR #239, opened 2026-03-16
238 periodically send emergency stop message over canpb

mlecriva:
## Summary

- **Periodic resend of emergency stop & fault statuses over CAN**: The power supply GPIO thread now wakes on a configurable timer (default 100 ms) instead of blocking on ISR-only events, ensuring receivers that missed an initial message or started late still learn about active fault conditions.
- **Shared PB_EmergencyStopStatus proto**: Moved the emergency stop protobuf message from `pf-power-supply` to `pf-common` so that `pf-robot-motion-control` and `pf-robot-motors` can subscribe to it.
- **Fixed inverted boolean semantics**: Renamed the proto field from `emergency_stop` to `emergency_stop_engaged` and inverted the GPIO mapping so that `true` means the stop is actually engaged, eliminating a confusing inversion.
- **Emergency stop subscription on motor & motion-control platforms**: Both `pf-robot-motors` and `pf-robot-motion-control` now register a CAN handler for the emergency stop message: motors disables all actuators, motion-control triggers a full game-end reset.
- **Missing game_end reset for TRACKER_SPEED_TUNING**: Added the missing case in `pf_handle_game_end` so the speed tuning controller is properly reset between game cycles.


gdoffe (changes requested):
LGTM ! Small missing new line at end of file.

gdoffe (approved):

PR #240, opened 2026-03-24
213 add settings partition to store key value pair

### LINKED ISSUE #213: Add settings partition to store key value pair
### Need

  Store configuration parameters as key-value pairs to persist settings between reboots.

  **Requirements:**
  - Persist application settings across device reboots
  - Load default values when flash is empty (first boot)
  - Work on both `cogip_native` (host) and `cogip_board` (STM32 hardware)
  - Provide a test application demonstrating the feature
  - Optional shell interface to browse and modify settings at runtime

### Solution: FlashDB (https://doc.riot-os.org/group__pkg__flashdb.html)

  **Why FlashDB?**
  - **Native KV storage**: Real key-value API (not file-based simulation)
  - **Minimal overhead**: Direct flash access via MTD, no filesystem required
  - **Auto-initialization**: Default values loaded automatically on first boot
  - **Dual platform support**: Works on STM32 flash (`mtd_flashpage`) and native RAM (`mtd_emulated`)
  - **protected access**: Thread-safe, wear-leveling, corruption resistant
  - **Small footprint**: Perfect for < 20 settings parameters

  **Storage backend can be initialised at board level**
  - STM32 hardware: Internal flash via `mtd_flashpage` (last flash pages)
  - Native platform: RAM-emulated MTD for testing

mlecriva:
## Summary

Add persistent key-value parameter storage using FlashDB, enabling motion control parameters to survive reboots without manual reconfiguration.

### Flash storage backend
- Integrate FlashDB KVDB as a singleton (`FlashKVStorage`) with mutex-based thread safety and typed `store<T>`/`load<T>` templates
- Configure MTD backends for both cogip-board (STM32 internal flash, last 8 pages / 16 KB) and cogip-native (RAM-emulated)
- Define a `settings` FAL partition on both boards
- Initialize the flash KV backend in `pf-common` so all applications get persistence automatically

### Parameter policy system
- Add `WithFlashStorage<KeyHash>` policy that persists values to flash on commit and restores them on init, using FNV-1a hashed keys
- Introduce `on_commit` / `on_init` policy trait detection with fold expressions in `PolicyTraits.hpp`
- Add `on_pb_read` / `on_pb_copy` protobuf conversion hooks with two concrete policies: `SpeedConversion` (units/s ↔ units/period) and `AccelerationConversion` (units/s² ↔ units/period²)
- Add virtual `load()` to `ParameterBase` for runtime reload without knowing the concrete type

### Motion control platform
- Centralize all `Parameter<>` declarations, key hashes, and policy composition in `motion_control_parameters.hpp` — robot config headers now only provide `constexpr` defaults
- Attach `WithFlashStorage` to all non-read-only parameters (PID gains, pose thresholds)
- Register quadpid and tracker chain PID gains in the CAN protobuf registry (fixes previous incorrect mapping)
- Add speed/acceleration limit and pose threshold parameters with transparent unit conversion at the protobuf boundary
- Load persisted values at platform init before motors and controllers start
- Remove unused `PidEnum` and related types

### Example
- Extend `parameter_demo` to exercise flash store/load/delete with the `WithFlashStorage` policy

### DIFF
--- a/boards/cogip-board/include/mtd_conf.h
+++ b/boards/cogip-board/include/mtd_conf.h
@@ -1,0 +1,75 @@
+/*
+ * Copyright (C) 2026 COGIP
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     boards_cogip_board
+ * @{
+ *
+ * @file
+ * @brief       MTD configuration for persistent settings storage
+ *
+ * This file defines the MTD (Memory Technology Device) configuration
+ * for persistent key-value storage using FlashDB on the cogip-board.
+ *
+ * @author      Mathis LECRIVAIN <lecrivain.mathis@gmail.com>
+ */
+
+#ifndef MTD_SETTINGS_H
+#define MTD_SETTINGS_H
+
+#include "mtd_flashpage.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief MTD Memory Technology Device interface configuration for settings storage
+ *
+ * Configuration for persistent key-value storage using FlashDB.
+ * Uses the last pages of internal flash memory (STM32G474RE).
+ *
+ * We cannot use MTD_FLASHPAGE_INIT_VAL() because it sets sector_count
+ * to FLASHPAGE_NUMOF (= entire flash, 256 sectors / 512 KB), which would
+ * expose the firmware code region to FlashDB. Instead we manually limit
+ * the device to 8 pages (16 KB).
+ *
+ * Memory layout (last 8 pages of flash):
+ * - Offset: FLASHPAGE_NUMOF - 8 = page 248 (on STM32G474RE)
+ * - Sector count: 8
+ * - Pages per sector: 1
+ * - Page size: 2048 bytes (hardware flash page size)
+ * - Total size: 8 × 2048 = 16 KB
+ *
+ * @{
+ */
+#define MTD_SETTINGS_SECTOR_COUNT (8)
+#define MTD_SETTINGS_PAGES_PER_SECTOR (1)
+#define MTD_SETTINGS_PAGE_SIZE (FLASHPAGE_SIZE)
+#define MTD_SETTINGS_OFFSET (FLASHPAGE_NUMOF - MTD_SETTINGS_SECTOR_COUNT)
+#define MTD_SETTINGS_INIT_VAL                                                                      \
+    {                                                                                              \
+        .base =                                                                                    \
+            {                                                                                      \
+                .driver = &mtd_flashpage_driver,                                                   \
+                .sector_count = MTD_SETTINGS_SECTOR_COUNT,                                         \
+                .pages_per_sector = MTD_SETTINGS_PAGES_PER_SECTOR,                                 \
+                .page_size = MTD_SETTINGS_PAGE_SIZE,                                               \
+                .write_size = 1,                                                                   \
+            },                                                                                     \
+        .offset = MTD_SETTINGS_OFFSET,                                                             \
+    }
+
+/** @} */
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* MTD_SETTINGS_H */
+/** @} */
--- a/lib/parameter/include/parameter/Parameter.hpp
+++ b/lib/parameter/include/parameter/Parameter.hpp
@@ -58,7 +58,7 @@ template <typename T, typename... Policies> class Parameter : public ParameterIn
 
   public:
     /// @brief Default constructor - initializes with invalid state
-    Parameter() : value_{}, valid_(false), changed_(true), mutex_(MUTEX_INIT)
+    Parameter() : value_{}, default_value_{}, valid_(false), changed_(true), mutex_(MUTEX_INIT)
     {
         mutex_init(&mutex_);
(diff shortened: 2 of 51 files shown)


lib/parameter/include/parameter/Parameter.hpp
```diff
@@ -82,11 +82,37 @@ template <typename T, typename... Policies> class Parameter : public ParameterIn
         valid_ = combined_on_set<T, Policies...>(temp);
         if (valid_) {
             value_ = temp;
+            combined_on_commit<T, Policies...>(value_);
```
gdoffe: `combined_on_commit` fires while the Parameter mutex is still held. With `WithFlashStorage`, that means a `FlashKVStorage::store()` call (an STM32 flash erase/program cycle, typically several ms) blocks every concurrent `get()` on the same Parameter. For PID gains read every 20 ms motion-control tick, one `set()` from CAN can stall the control loop for a whole cycle.

Drop the mutex around the commit (`mutex_unlock(&mutex_); combined_on_commit<T, Policies...>(value_); return valid_;`, with a snapshot of `value_` taken before unlocking to keep the commit thread-safe against a subsequent `set()`).

platforms/pf-robot-motion-control/include/motion_control_parameters.hpp
```diff
@@ -4,32 +4,235 @@
...
+
+#include "KeyHash.hpp"
+#include "app_conf.hpp"
 #include "canpb/ReadBuffer.hpp"
+#include "parameter/ConversionPolicies.hpp"
+#include "parameter/Parameter.hpp"
+#include "parameter/StoragePolicies.hpp"
+
+using namespace cogip::parameter;
+using cogip::utils::operator"" _key_hash;
```
gdoffe, line 23: `using namespace cogip::parameter;` at file scope in a public header leaks the whole `cogip::parameter` namespace into every translation unit that transitively includes this file (which, with `#include "motion_control.hpp"` chaining in, is basically every .cpp in the platform). Move the using-directive inside the `cogip::pf::motion_control` namespace block below (line 107-147), or qualify the Parameter declarations with `cogip::parameter::` explicitly. I personnaly prefer the last one.

boards/cogip-board/include/mtd_conf.h
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * Memory layout (last 8 pages of flash):
+ * - Offset: FLASHPAGE_NUMOF - 8 = page 248 (on STM32G474RE)
+ * - Sector count: 8
+ * - Pages per sector: 1
+ * - Page size: 2048 bytes (hardware flash page size)
+ * - Total size: 8 × 2048 = 16 KB
+ *
+ * @{
+ */
```
gdoffe, line 50: `MTD_SETTINGS_OFFSET = FLASHPAGE_NUMOF - 8` carves the last 16 KB of STM32G474RE flash for KV storage, but nothing in `boards/cogip-board/Makefile.include` or a linker script restricts the firmware image from growing into that region. Once `.text + .rodata + .data` grows past `FLASH - 16K`, the flasher happily overwrites the settings, and the first `FlashKVStorage::store()` after boot erases a sector containing firmware ==>> brick. Add `ROM_LEN` (or `IMAGE_SIZE_MAX`, whichever RIOT's stm32.inc.mk honours on this board) to cap the image at `FLASHPAGE_NUMOF - 8` pages, and fail the build if the image exceeds that.

platforms/pf-common/platform_common.cpp
```diff
@@ -114,6 +115,13 @@ int pf_init(copilot_callback_t on_connect, copilot_callback_t on_disconnect)
     on_copilot_connected_callback = on_connect;
     on_copilot_disconnected_callback = on_disconnect;
 
+    // Initialize flash key-value storage
+    int flash_res = flash_kv_storage::FlashKVStorage::instance().init();
+    if (flash_res) {
+        LOG_ERROR("Flash KV storage initialization failed, error: %d\n", flash_res);
+        return flash_res;
```
gdoffe: Returning the error here aborts `pf_common::pf_init` without ever reaching `canpb.init(...)` below. The caller on pf-robot-motion-control (see `platforms/pf-robot-motion-control/platform.cpp:36-83`) treats the non-zero return as "skip all platform-specific handler registrations" but still calls `pf_init_motion_control()` right after, so the robot boots with motors initialised and zero CAN handlers, meaning no one can stop it. FlashKVStorage already returns an error silently for every `store`/`load`/`del` while not initialised, so this path is safe for runtime: log the failure (you already do) and keep going instead of returning.

gdoffe:
Few things but good job overall. take care of coding rules.

gdoffe (changes requested):

platforms/pf-common/platform_common.cpp
mlecriva: Fixed

boards/cogip-board/include/mtd_conf.h
mlecriva: Fixed

platforms/pf-robot-motion-control/include/motion_control_parameters.hpp
mlecriva: Fixed

lib/parameter/include/parameter/Parameter.hpp
mlecriva: Fixed.

gdoffe (approved):
good job !

PR #242, opened 2026-04-12
lift safety fixes and brake chain mechanism

gdoffe:
## Summary

Two threads bundled on this branch:

**Lift safety fixes**
- Always recalibrate lift position when the lower limit switch triggers, not only during the init sequence.
- Drop the `actuate()` call from the lower limit switch handler: only recalibrate and disable, so the motor is not re-enabled after a controlled stop.
- On the upper limit switch, latch the engine brake chain to actively hold the lift against gravity (a passive DRV8873 hardware brake alone cannot hold the load). A freeze-target implementation is kept as a working reference in an intermediate commit before the final switch-over to the brake chain.
- `pf-robot-motors`: do not re-enable motors on game reset / game start (they may have been disabled by the user for a reason).
- `pf-robot-motors`: `actuator_init` CAN message now accepts an optional `PB_ActuatorInit { optional PB_PositionalActuatorEnum id }` payload; empty payload keeps the legacy init-all behaviour, a set id initialises only that actuator.

**Brake chain mechanism**
- `BaseControllerEngine` gains a brake latch and a brake controller chain: when the latch is set, the engine runs the brake chain instead of the normal one and still calls `process_outputs`, so the brake chain's speed command flows through to the motor driver. The latch stays set until an explicit `set_brake(false)`.
- `Motor` builds a minimal brake chain (ResetController forcing `speed_order = 0` + dedicated SpeedPIDController) and registers it on its MotorEngine. Any `Motor`-based actuator gets active-hold brake for free. `Motor::actuate()` explicitly releases the brake so a new motion request resumes normally; other mutators (`set_target_distance`, etc.) keep it latched.
- `pf-robot-motion-control` builds a brake chain for the platform engine (ResetController zeroing both polar speed orders + PolarParallel of dedicated linear and angular SpeedPIDControllers).
- The mechanism is distinct from EMS: EMS still disables motors (cuts power) and is not wired through the brake chain.

## Context

Initial trigger: with the emergency stop engaged at boot, the lift init sequence timed out and disabled the motor. When the button was released the lower limit switch fired (gravity), the handler called `actuate(0)` which re-enabled the motor with an uncalibrated 130 mm offset, producing a violent downward lunge.

While addressing that, we also saw the lift slip down at the upper limit because a passive hardware brake cannot counteract gravity, and found that `actuator_init` could not target a single lift of a dual-lift setup. The brake chain mechanism generalises active-stop / active-hold into a reusable pattern ready to be wired up for controlled robot stops (trigger left to policy; EMS keeps its current disable behaviour).

### DIFF
--- a/lib/actuator/Lift.cpp
+++ b/lib/actuator/Lift.cpp
@@ -26,6 +26,13 @@ Lift::Lift(const LiftParameters& params)
 
 void Lift::init()
 {
+    // Invalidate any stale command from a previous run: init can be
+    // re-triggered over CAN, and the "else" branch below (already at
+    // lower limit) never calls actuate(), so last_command_ would
+    // otherwise keep a value from a previous session and cause the
+    // next matching actuate() to be wrongly skipped.
+    last_command_ = INT32_MIN;
+
     int ret =
         LiftsLimitSwitchesManager::instance().register_gpio(params_.lower_limit_switch_pin, this);
     if (ret) {
@@ -123,6 +130,9 @@ void Lift::at_lower_limit()
         LOG_INFO("Lower limit switch pressed\n");
         set_current_distance(params_.lower_limit_mm);
         disable();
+        // Invalidate last_command_ so the next actuate() with the same
+        // target is not skipped and can re-enable the motor.
+        last_command_ = INT32_MIN;
     } else {
         LOG_INFO("Lower limit switch released\n");
     }
--- a/lib/actuator/Lift.cpp
+++ b/lib/actuator/Lift.cpp
@@ -133,8 +133,14 @@ void Lift::at_upper_limit()
     // Only react when switch is pressed (reads 1 with active-high logic)
     if (gpio_read(params_.upper_limit_switch_pin)) {
         LOG_INFO("Upper limit switch pressed\n");
-        motor_engine_.set_target_speed(0);
         motor_engine_.set_timeout_enable(false);
+        // Hold the current position. A hardware brake alone would only short
+        // the stator (dynamic braking) and cannot hold the lift against
+        // gravity, so freeze the target at the current distance and let the
+        // PID actively compensate.
+        const float current = motor_engine_.get_current_distance_from_odometer();
+        motor_engine_.set_target_distance(current);
+        last_command_ = static_cast<int32_t>(current);
     } else {
         LOG_INFO("Upper limit switch released\n");
     }
--- a/lib/actuator/Lift.cpp
+++ b/lib/actuator/Lift.cpp
@@ -122,8 +122,7 @@ void Lift::at_lower_limit()
     if (gpio_read(params_.lower_limit_switch_pin)) {
         LOG_INFO("Lower limit switch pressed\n");
         set_current_distance(params_.lower_limit_mm);
-        actuate(params_.lower_limit_mm);
-        motor_engine_.set_timeout_enable(false);
+        disable();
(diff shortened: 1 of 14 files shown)


lib/actuator/Lift.cpp
```diff
@@ -133,8 +133,16 @@ void Lift::at_upper_limit()
     // Only react when switch is pressed (reads 1 with active-high logic)
     if (gpio_read(params_.upper_limit_switch_pin)) {
         LOG_INFO("Upper limit switch pressed\n");
-        motor_engine_.set_target_speed(0);
         motor_engine_.set_timeout_enable(false);
+        // Hold the current position. A hardware brake alone would only short
```
mlecriva: In commit message, I think you can remove this part:

> In
> DUALPID_TRACKER mode this regenerates a zero-distance profile, taking
> tracker_velocity to zero while the feedback loop holds the position.
> 

And maybe same in the comment, WDYT?

motion_control/motion_control_common/BaseControllerEngine.cpp
```diff
@@ -58,7 +58,12 @@ void BaseControllerEngine::thread_loop()
         // Clear modified flags
         io_.clear_modified();
 
-        if ((enable_) && (controller_)) {
+        if (brake_ && brake_controller_) {
```
mlecriva: It this a design choice to allow to execute braking behaviour without checking for `enable_` ? I think this can cause some trouble, WDYT?

mlecriva (changes requested):

lib/actuator/Lift.cpp
gdoffe: Commit message reworded and the inline comment cleaned in the same commit (now `8882c35b` after the rebase). The later commit on this branch (`lib/actuator/lift: latch the engine brake on upper limit switch`) still replaces the block entirely with the brake-chain version, so the final state of `Lift.cpp` is unchanged (only the intermediate commit no longer mentions `DUALPID_TRACKER` / `tracker_velocity`).

motion_control/motion_control_common/BaseControllerEngine.cpp
gdoffe: Good point, applied. I hesitated at first on whether brake, being a safety-ish mechanism, should take priority over `enable_`,  but you're right, disable is disable. The thread loop now gates the brake branch on `enable_` as well:

So after a disable (EMS or timeout) the brake chain stops running until the engine is re-enabled. The explicit unbrake policy (`set_brake(false)` only) is unchanged.

motion_control/motion_control_common/BaseControllerEngine.cpp
```diff
@@ -58,28 +58,34 @@ void BaseControllerEngine::thread_loop()
...
+
+                    enable_ = false;
+                }
+
+                // Process controller outputs
+                process_outputs();
             }
-
-            // Process controller outputs
-            process_outputs();
```
mlecriva, line 82: process_outputs();  can stay here in order to duplicate code inside if/else branches

mlecriva (changes requested):
One last thing

motion_control/motion_control_common/BaseControllerEngine.cpp
gdoffe: done

mlecriva (approved):
LGTM

PR #243, opened 2026-04-20
241 add SparkFun OTOS optical tracking sensor support

### LINKED ISSUE #241: Add SparkFun OTOS optical tracking odometry sensor support
## Context

The current odometry relies on wheel encoders (`OdometerEncoder`) which suffer from
wheel slippage, especially during fast accelerations or when the robot is pushed.
The SparkFun Optical Tracking Odometry Sensor (OTOS, PAA5160E1) combines an optical
tracking sensor with a 6-axis IMU and onboard sensor fusion to provide accurate
ground-truth position (X, Y) and heading data over I2C, independent of wheel contact.

## Goal

Add a new `OdometerOTOS` class inheriting from `OdometerInterface` to provide an
alternative odometry source for the motion control platform.

## Key features

- I2C driver for the OTOS sensor (PAA5160E1)
- `OdometerOTOS` class implementing `OdometerInterface` (init, update, distance_mm, delta_distance_mm, set_distance_mm)
- Support for dual-axis (X, Y) tracking and heading from onboard sensor fusion
- Configurable mounting offset and orientation correction
- Usable as drop-in replacement for `OdometerEncoder` in `MotorParameters`

## Hardware

- SparkFun Optical Tracking Odometry Sensor (Qwiic)
- Communication: I2C
- Datasheet: https://docs.sparkfun.com/SparkFun_Optical_Tracking_Odometry_Sensor/introduction/

## Implementation notes

- The OTOS provides fused 2D position + heading, but `OdometerInterface` is 1D (single distance).
  A `LocalizationOTOS` implementing `LocalizationInterface` (2D pose) may be more appropriate
  for the platform engine, while `OdometerOTOS` could project onto a single axis for lift use.
- Consider whether the sensor fusion should replace or complement the existing encoder-based
  localization.
- The I2C driver should be implemented as a RIOT module under `lib/` or `drivers/`.

gdoffe:
## Summary

Adds support for the SparkFun OTOS optical tracking odometry sensor as a drop-in alternative to wheel-encoder localization, and exposes its two calibration scalars through the CAN parameter system so they can be tuned at runtime.

**Driver and integration**
- `drivers/otos/`: new SparkFun OTOS I2C driver (register map, pose/velocity reads, scalar and offset write, IMU calibration).
- `drivers/soft_i2c/`: software I2C bit-banging driver, used because the OTOS chip sits on pins that conflict with the on-chip I2C peripheral on the robot6 wiring.
- `lib/localization/`: new `LocalizationOTOS` sub-module implementing `LocalizationInterface` on top of the OTOS driver. Computes the polar delta expected by the motion control loop from the absolute pose reported by the sensor.
- `lib/localization/LocalizationInterface`: promoted `init`, `reset` and `send_telemetry` to the interface so that PlatformEngine does not need to know which concrete localization is plugged in.
- `applications/robot-motion-control`: `robot_localization` is now defined per robot configuration; `robot6_conf.hpp` wires the OTOS path, other robots keep the differential encoder path.
- `examples/otos_test/`: standalone app for bring-up, reads and prints the OTOS pose in a loop.

**Runtime calibration plumbing**
- `LocalizationOTOS::apply_scalars(linear, angular)` pushes new scalars to the sensor chip and caches them locally.
- `robot6_conf.hpp` switches `otos_linear_scalar` and `otos_angular_scalar` from `constexpr` to live `Parameter<float>` and defines a `ROBOT_HAS_OTOS` marker.
- `platforms/pf-robot-motion-control/motion_control_parameters.cpp` registers the two scalars in the CAN parameter handler and, on a successful set, calls `pf_motion_control_apply_otos_scalars()` so the chip picks up the new calibration without a reboot.
- Robots 1–5 are unaffected (the whole OTOS block is gated by `ROBOT_HAS_OTOS`).
- Offsets remain `constexpr`: they describe sensor-to-robot mounting geometry, not a runtime calibration.

**Drive-by build fix**
- `drivers/otos/OTOS.cpp`: drop an unused `update_count` local that tripped `-Werror=unused-variable` on the `ROBOT_ID=6` build.

## Context

Wheel-encoder odometry on this robot suffers from slip, particularly during aggressive moves. The SparkFun OTOS (PAA5160E1 optical flow + LSM6DSO IMU, on-board STM32 fusion) gives a ground-truth 2D pose (x, y, heading) via I2C independently of wheel contact, which we want as a drop-in replacement.

To make the sensor useful in practice we also need to calibrate its two scalars per robot without going through a recompile/flash cycle; the host-side tool for that lives on the `cogip-tools` side (`cogip/tools/firmware_otos_calibration/` on branch `add-firmware-otos-calibration-tool`) and drives the parameters exposed here.

## Branching

This branch is stacked on top of `fix/lift-emergency-stop-behavior` (#242). Review / merge that one first; the OTOS commits start at `7af4356c` and are strictly additive (plus the `ROBOT_HAS_OTOS` gating in `pf-robot-motion-control`).


drivers/otos/include/otos/OTOS.hpp
```diff
@@ -0,0 +1,128 @@
+// Copyright (C) 2025 COGIP Robotics association <cogip35@gmail.com>
```
mlecriva: 2025 -> 2026 on every files

lib/localization/localization_otos/LocalizationOTOS.cpp
```diff
@@ -0,0 +1,93 @@
...
+    if (first_update_) {
+        polar_.set_distance(0.0f);
+        polar_.set_angle(0.0f);
+        first_update_ = false;
+        return 0;
+    }
+
+    // Compute linear delta by projecting displacement onto heading direction
+    // (matches the convention of LocalizationDifferential where positive = forward)
+    float dx = pose_.x() - prev_pose_.x();
```
mlecriva, line 114: The OTOS driver gives a speed getter API, I there a reason for not using it here ?

mlecriva: Oh, maybe it's because you can only get the speed along X and Y axis and not the angular speed ?

platforms/pf-robot-motion-control/Makefile.dep
```diff
@@ -43,9 +43,9 @@ USEMODULE += telemetry_controller
 # Board features required
 FEATURES_REQUIRED += periph_pwm
 
-FEATURES_REQUIRED += periph_qdec
-
 # OTOS localization support for robots using optical tracking
 ifneq (,$(filter 6,$(ROBOT_ID)))
   USEMODULE += localization_otos
+else
```
mlecriva, line 49: This should probably be moved in the commit where you add `USEMODULE += localization_otos`

drivers/otos/OTOS.cpp
```diff
@@ -71,8 +71,6 @@ int OTOS::init()
 
 int OTOS::update()
 {
-    static uint32_t update_count = 0;
```
mlecriva: This could be squashed with the OTOS driver addition commit, WDYT?

applications/robot-motion-control/include/robot6_conf.hpp
```diff
@@ -136,21 +136,28 @@ inline Parameter<float, NonNegative> tracker_angular_speed_pid_integral_limit{
...
+// mounting geometry, calibrated at assembly, no runtime tuning.
 constexpr float otos_offset_x_mm = 0.0f;
 constexpr float otos_offset_y_mm = 0.0f;
 constexpr float otos_offset_h_deg = 0.0f;
 
-static constexpr cogip::localization::LocalizationOTOS::Parameters otos_params = {
-    .linear_scalar = otos_linear_scalar,
-    .angular_scalar = otos_angular_scalar,
+static cogip::localization::LocalizationOTOS::Parameters otos_params = {
+    .linear_scalar = otos_linear_scalar.get(),
```
mlecriva: Here you can move to Direct `ParameterInterface` reference inside `LocalizationOTOS::Parameters` structure.

As done in PID lib for example. 

You can configure the OTOS scalar ranges like this:

```C++
Parameter<float, WithBounds<min,max>> otos_linear_scalar{...};

or

Parameter<float, Clamp<min,max>> otos_linear_scalar{...};
```

I think it is possible to remove the `apply_scalars` methods from the previous commit and let the Parameter backend do the job by adding a `has_changed()` API to the parameter Interface and move this internally to the OTOS driver.

I would prefer to keep `void pf_handle_parameter_set(cogip::canpb::ReadBuffer& buffer)` as generic as possible and avoid adding specific cases handling inside.

We can discuss it synchronously.

mlecriva: I think the `has_changed` API will be also useful for other parameters like speed and acceleration

mlecriva (changes requested):
Really nice job !

drivers/otos/include/otos/OTOS.hpp
gdoffe: Done — all new files added by this PR updated from 2025 to 2026 in their introducing commit. Pre-existing files (LocalizationInterface, LocalizationDifferential, motion_control_parameters.cpp) kept their original year since they are not added here.

lib/localization/localization_otos/LocalizationOTOS.cpp
gdoffe: Yes, exactly. The OTOS velocity API only reports linear velocity along X and Y (world frame); it does not expose an angular/heading rate. Since the motion control loop feedback needs a polar delta (linear along heading + angular), we still have to derive the angular delta from `pose(t) - pose(t-1)`. Using the driver's linear velocity getter for the linear component alone would just couple two different integration paths (a discrete diff for one, the chip filter for the other), so it feels cleaner to read both from the pose delta.

platforms/pf-robot-motion-control/Makefile.dep
gdoffe: Done — the `else/FEATURES_REQUIRED += periph_qdec` block is now in the commit that adds `USEMODULE += localization_otos` (`applications/robot-motion-control: move localization to robot config`). The subsequent soft_i2c switch commit no longer touches `Makefile.dep`.

drivers/otos/OTOS.cpp
gdoffe: Done — the unused `update_count` was never introduced: the fix is squashed into the OTOS driver addition commit, so the separate "drop update_count" commit is gone.

applications/robot-motion-control/include/robot6_conf.hpp
gdoffe: Applied, thanks — this is much cleaner. The stack ends with two new commits squashed into the OTOS sequence:

- `lib/parameter: add has_changed / clear_changed API to ParameterInterface`: new virtual pair on `ParameterInterface<T>`; the flag is set on every successful `set()` and is `mutable` so a consumer holding a `const ParameterInterface<T>&` can reset it after handling the value (same ownership pattern as `PIDParameters`). Newly constructed parameters start as `changed=true` so the very first poll picks up the initial value.
- `localization/otos: react to scalar changes via the parameter polling API`: `LocalizationOTOS::Parameters` now holds `const ParameterInterface<float>&` for both scalars (floats kept only for the mounting offsets). `update()` polls `has_changed()` on each scalar and re-programs the OTOS chip when either has changed; `init()` pushes the initial values and clears the flag.

`pf_handle_parameter_set` stays fully generic — no OTOS-specific branch and no `pf_motion_control_apply_otos_scalars()` helper (the intermediate `apply_scalars()` commit is dropped, it was never needed in the final story). The `has_changed()` API is on the base interface so other consumers (speed / acceleration filters, etc.) can reuse it exactly as you suggested.

mlecriva (approved):
LGTM good job

PR #245, opened 2026-04-22
motion_control: robust new-target detection and clean end-of-motion

gdoffe:
## Summary

- Centralize new-target detection in `TargetChangeDetector` with epsilon-based
  float comparison (`areFloatsEqual`) and support for 5 watched keys including
  `motion_direction` and `bypass_final_orientation`. A pose order that differs
  from the previous one only on a flag (same x/y/O but different
  `bypass_final_orientation`) is now correctly seen as a new command.
- Drop the hand-rolled (x, y, O) comparison in `PoseStraightFilter` and have
  the filter consume the detector's `new_target` IO key. Remove the sticky
  `bypass_final_orientation_` cache that caused FINISHED-without-rotation on
  some motions (robot ending at -96° instead of 180°).
- `PoseStraightFilter::reset()` no longer touches the state machine. The
  TargetChangeDetector is the sole authority on re-arming the state machine
  to `ROTATE_TO_DIRECTION`, so the cosmetic
  `ROTATE_TO_DIRECTION → MOVE_TO_POSITION` cycle that used to fire on every
  successful `reached` to settle a few mm of drift is gone.
- Skip `quadpid_tracker_chain::reset()` on `reached` / `intermediate_reached`
  (Mathis's fix). The cascade now only fires on genuinely aborted motions
  (`blocked`, `timeout`), so the PIDs keep their integrals across a clean
  arrival and the next motion starts from a warm state.

Rename `areDoublesEqual` → `areFloatsEqual` as a cleanup (the function has
always taken `float`).

### DIFF
--- a/motion_control/filters/path_manager_filter/PathManagerFilter.cpp
+++ b/motion_control/filters/path_manager_filter/PathManagerFilter.cpp
@@ -101,6 +101,16 @@ void PathManagerFilter::execute(ControllersIO& io)
     // Emit bypass_final_orientation from current waypoint
     if (!keys_.bypass_final_orientation.empty()) {
         io.set(keys_.bypass_final_orientation, current->bypass_final_orientation());
+        // [DEBUG bypass_final_orientation] what is being written to IO
+        static bool prev_bypass = false;
+        bool cur_bypass = current->bypass_final_orientation();
+        if (cur_bypass != prev_bypass) {
+            LOG_INFO("PMF: bypass_final_orientation -> %d (wp=%u/%u)\n",
+                     static_cast<int>(cur_bypass),
+                     static_cast<unsigned>(path.current_index() + 1),
+                     static_cast<unsigned>(path.size()));
+            prev_bypass = cur_bypass;
+        }
     }
 
     // Emit motion_direction from current waypoint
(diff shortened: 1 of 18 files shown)


motion_control/filters/pose_straight_filter/PoseStraightFilter.cpp
```diff
@@ -142,6 +142,9 @@ void PoseStraightFilter::execute(ControllersIO& io)
     }
 
     if (new_target) {
+        LOG_INFO("Recompute triggered: target=(%.2f, %.2f, %.2f)",
```
mlecriva: Maybe we can drop this commit as it was initially added for debug purpose, WDYT ?

motion_control/filters/path_manager_filter/PathManagerFilter.cpp
```diff
@@ -101,6 +101,16 @@ void PathManagerFilter::execute(ControllersIO& io)
     // Emit bypass_final_orientation from current waypoint
     if (!keys_.bypass_final_orientation.empty()) {
         io.set(keys_.bypass_final_orientation, current->bypass_final_orientation());
+        // [DEBUG bypass_final_orientation] what is being written to IO
```
mlecriva: Same for this commit, I think this is not relevant to merge this kind of change

mlecriva (changes requested):

motion_control/filters/pose_straight_filter/PoseStraightFilter.cpp
gdoffe: Done, commit dropped via rebase.

motion_control/filters/path_manager_filter/PathManagerFilter.cpp
gdoffe: Done, commit dropped via rebase.

mlecriva (approved):
LGTM

PR #246, opened 2026-04-24
235 print firmware version inside main app

mlecriva:
build: define MCU_FIRMWARE_BUILD_VERSION_STRING from git describe
- Needed to display firmware version from application logs and identify
running firmware on each board
- Overridable via MCU_FW_BUILD_VERSION for tagged releases
- Convenience make target print-build-fw-version for scripts

applications: log firmware version at startup
- Helps identify which firmware is running on each board
  during debugging and field diagnostics without needing
  to reflash
- Motion control also logs ROBOT_ID to distinguish boards
  in multi-robot setups


ecourtois (approved):
LGTM after testing!

PR #247, opened 2026-04-25
Fix lift init wedge, switch direction, and dedicate a brake PID

gdoffe:
## Summary

- **Wake homing wait on switch hit**: replace the unconditional 2 s `ztimer_sleep` in `Lift::init()` with a mutex acquired up front and released by `at_lower_limit()` on the rising edge. The init returns the moment homing is physically done; the CAN handler thread no longer wedges on every init.
- **Newline on `_handle_command` logs**: add the missing trailing `\n` on the three logs at the tail of `_handle_command` so concurrent ISR notices stop being concatenated to them in the `cogip-mcu-logger` stream.
- **Direction-gated limit-switch handling**: only ignore a switch press when the lift is genuinely above (resp. below) the switch and moving away from it. Overshoot through a switch is still honoured as a genuine end-of-travel; bounces during the opposite-direction motion no longer freeze an in-flight command.
- **Dedicated brake PID**: split the brake chain off the tracking PID with its own `PIDParameters` and `PID` instance. Values match the tracking gains, so behaviour is unchanged out of the box; the integrator state is now isolated, unblocking future independent tuning of the brake against gravity without touching trajectory tracking.

### DIFF
--- a/applications/robot-lift-control/include/lift_common_conf.hpp
+++ b/applications/robot-lift-control/include/lift_common_conf.hpp
@@ -50,6 +50,20 @@ inline cogip::parameter::Parameter<float, cogip::parameter::NonNegative> motor_l
 inline cogip::parameter::Parameter<float, cogip::parameter::NonNegative>
     motor_lift_speed_pid_integral_limit{static_cast<float>(etl::numeric_limits<int16_t>::max())};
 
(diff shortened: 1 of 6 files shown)


mlecriva (approved):
LGTM

PR #248, opened 2026-05-03
Configure robot2 for the 2026 Ninja drivetrain

gdoffe:
## Summary

- **Retune robot2 for 25GA-370 motors**: the GA32Y-31ZY drive motors had to be dropped for the 25GA-370 12 V / 500 RPM-rated variant to fit under the Eurobot weight cap. The new chain (5952 RPM motor shaft, 9.6:1 internal gearbox, 20T/44T timing belt) only spins the wheel at 281.8 RPM no-load against the previous 1544 RPM, so the existing motor_constant and kinematic caps would saturate the PWM the moment the controller asks for anything close to the old 2000 mm/s ceiling. Recompute motor_constant on the new no-load wheel speed and pull the linear and angular caps down to ~80% of the achievable wheel velocity so the speed PID keeps PWM headroom to correct.
- **Recalibrate robot2 OTOS linear scalar**: bench measurement showed the OTOS reporting 800 mm for an actual 809 mm travel, so the previous scalar of 0.969 was undershooting. Bump it to 0.976080 to bring reported distance back in line with the physical measurement.

### DIFF
--- a/applications/robot-motion-control/include/robot2_conf.hpp
+++ b/applications/robot-motion-control/include/robot2_conf.hpp
@@ -171,7 +171,7 @@ constexpr float default_brake_angular_speed_pid_integral_limit =
 constexpr uint8_t otos_i2c_addr = 0x17;
 
 // OTOS calibration scalar defaults (range [0.872, 1.127]).
-constexpr float default_otos_linear_scalar = 0.969f;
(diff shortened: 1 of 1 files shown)


mlecriva (approved):
LGTM

PR #249, opened 2026-05-03
PlatformEngine: latch the brake chain on reached arrivals

gdoffe:
## Summary

- **Latch brake on reached**: once the platform engine reported a pose as reached, the normal control chain kept running and `process_outputs` kept firing reached callbacks every cycle. Holding the robot at the target was left to the tracking PID, which has zero pose Ki by construction (the cascade can't afford a second integrator) and silently drifts on slope, friction, or when another robot pushes against it. Latch the dedicated brake chain on the rising edge of the final `reached` status: from the next cycle on, `BaseControllerEngine` runs the `ZeroSpeedOrder + SpeedPID` brake chain instead of the tracking chain, actively driving the robot toward zero speed. While `brake_` is set, `process_outputs` skips pose_reached tracking, transition detection, logging and the callback dispatch entirely. `intermediate_reached` and `blocked` are left untouched so path navigation and host-level blocked handling keep their current semantics. Releasing the brake on a new motion is the API caller's job, not the engine setter's: `pf_handle_target_pose`, `pf_handle_speed_order`, `pf_handle_start_pose` and `pf_handle_path_start` now call `set_brake(false)` alongside the pre-existing reset of pose_reached so the next motion runs the normal control chain again.
- **Dedicate brake PIDParameters**: the brake speed PIDs ran their own integrator state but borrowed their gains from `quadpid_chain::*_speed_pid_parameters`, so any tuning of the brake loop also moved the tracking loop. Add `brake_*_speed_pid_kp/ki/kd/integral_limit` Parameters in robot1 and robot2 conf, initialised to the same values as the existing speed loop so behaviour is unchanged out of the box. `brake_chain.hpp` now builds its `PIDParameters` from those new Parameters, drops the include of `quadpid_chain.hpp`, and binds the `brake_*_speed_pid` PID instances on the dedicated parameters.

### DIFF
--- a/applications/robot-motion-control/include/robot1_conf.hpp
+++ b/applications/robot-motion-control/include/robot1_conf.hpp
@@ -64,6 +64,21 @@ constexpr float default_tracker_angular_speed_pid_kp = 5.5;
 constexpr float default_tracker_angular_speed_pid_ki = 0.6;
 constexpr float default_tracker_angular_speed_pid_kd = 0;
 
+// ============================================================================
+// Brake chain PID gains (dedicated, runs while engine.brake_ is latched).
+// Initialised to the tracking values so behaviour is unchanged out of the
+// box; tuneable independently to fight static disturbances on hold.
+// ============================================================================
+
+// Linear speed PID (brake chain)
+constexpr float default_brake_linear_speed_pid_kp = 3.;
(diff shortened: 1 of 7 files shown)


mlecriva (approved):
LGTM

PR #251, opened 2026-05-06
applications/robot-motion-control: add brake PID config for robots 3-6

mlecriva:
- Robot 2 already had dedicated brake-chain PID gains since the brake chain was split from the tracker speed loop; robots 3, 4, 5 and 6 were left without them and fell back to defaults.
- Mirror each robot's tracker speed gains into new brake gains so hold behaviour stays identical out of the box.
- Derive matching integral limits from the brake gains, opening the door to tune the brake chain independently against static disturbances without touching tracking dynamics.


gdoffe (approved):
LGTM

PR #252, opened 2026-07-18
Cup/2026 work on site

gdoffe:
### Changes

- **Latch terminal states in MotorEngine:** pin pose_reached_ on the terminal (blocked/timeout) state until the next set_target_distance(), and skip the IO read while latched, so an upstream reached write can no longer leak through and fire a spurious success callback right after a failure. Error log moved inside the first-cycle guard to stop per-cycle spam.
- **Re-arm target change detector after reset:** a chain reset now sets first_run_ back to true so the next execute() refreshes cached values and re-arms new_target, even when the incoming target equals the last cached one (typical after a brake hold). Prevents the robot lurching forward on resume with a stale profile.
- **Engage brake chain instead of forcing filters finished:** blocked, manual-brake and game-end paths now call set_brake(true) so the platform engine swaps to the dedicated brake chain and actively drives motors to zero, replacing the old force_finished_state() plus hand-zeroing workarounds.
- **Add set_brake_locked() variant:** fixes the deadlock where the pose-reached callback (running under the engine mutex) called set_brake() and re-locked the same non-recursive mutex. Public set_brake() keeps its locking contract.
- **Reset brake controller on latch:** reset the brake controller on the false-to-true edge so the brake speed ramp captures the actual current speed instead of reusing stale state from the previous episode.
- **Clean brake state on disable:** release any latched brake and force both pose-straight filters to FINISHED on pf_disable_motion_control(), so re-enabling resumes the normal chain from a clean state.
- **Reset PIDs on early pose transition:** reset pose and speed PIDs when PoseStraightFilter jumps straight to ROTATE_TO_FINAL_ANGLE, so the anti-drift integral from the first rotation does not bias the final orientation.
- **Add reset to conditional switch meta-controller:** override the previously no-op reset() to propagate to both child branches, since either may be selected on the next execute().
- **Skip anti-blocking on lift ascent:** toggle the anti-blocking detector by move direction so slow gravity-loaded ascents are not mistaken for a stall; descents stay protected.
- **Fix speed PID integral windup:** size the tracker/brake integral caps in per_period units instead of per_s, removing the ~50x over-wind that caused start kicks and long bleed on direction reversal; limits replaced with empirically validated values.
- **Motion tuning:** cup-readiness retuning of robot 1, robot 2 (initial plus on-site round) and PAMI robots 3-6, with rebalanced tracker/brake PID gains, tightened pose thresholds, and revised speed/acceleration envelopes per chassis.
- **Cap PAMI speeds for homologation:** lower linear speed/acceleration limits on PAMI robots 3-6 so they move slowly and predictably during the homologation run.

### DIFF
--- a/applications/robot-motion-control/include/robot1_conf.hpp
+++ b/applications/robot-motion-control/include/robot1_conf.hpp
@@ -48,11 +48,11 @@ constexpr float default_angular_speed_pid_kd = 0;
 
 // Tracker linear pose PID (tracker during MOVE_TO_POSITION)
 // Ki helps eliminate steady-state error when tracker profile ends
-constexpr float default_tracker_linear_pose_pid_kp = 0.1;
+constexpr float default_tracker_linear_pose_pid_kp = 0.15;
 constexpr float default_tracker_linear_pose_pid_ki = 0.0;
 constexpr float default_tracker_linear_pose_pid_kd = 0;
 // Tracker angular pose PID (tracker during ROTATE states)
-constexpr float default_tracker_angular_pose_pid_kp = 0.3;
+constexpr float default_tracker_angular_pose_pid_kp = 0.15;
 constexpr float default_tracker_angular_pose_pid_ki = 0;
 constexpr float default_tracker_angular_pose_pid_kd = 0;
 // Tracker linear speed PID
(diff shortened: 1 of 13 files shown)


mlecriva (approved):
LGTM

